PR #7, opened 2025-07-28
Auto updating on server side install

0belous:
JSDelivr grabs the latest js file from github and proxies it with the correct headers, my code takes the file and puts it into a script element eliminating the need for manually changing code. This could probably be implemented for the userscript too but I don't use it so I'm not sure how it works.
```
(function() {
  'use strict';
  const scriptUrl = 'https://cdn.jsdelivr.net/gh/n00bcodr/Jellyfin-Enhanced@main/jf_enhanced.js';
  const script = document.createElement('script');
  script.src = scriptUrl;
  script.type = 'text/javascript';
  document.head.appendChild(script);
})();
```


n00bcodr:
I'd rather have this as another method mentioning that the script would be auto upgraded?

0belous:
sure thing, If I'm not mistaken you should be able to make edits and phrase it how you like

PR #137, opened 2025-10-28
Fix quality tags for TV mode

RoboMagus:
I noticed that on all devices I had the qualitytags visible, except for on TVs. 
Turns out that the `cardImageContainer` thats usually an `a href` type element, is a plain old `div` when TV display mode is used.

This PR removes the `a` from the queryselectors such that the `cardImageContainer`s can be found on TVs as well.


n00bcodr:
Thank you for the PR!

Good Catch, by that logic, genre tags and language tags also need fixing?

n00bcodr:
Yes they do!

I updated this PR with those, hope that's ok!

Merging it :)

n00bcodr:
This is now included in version [9.2.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/tree/9.2.0.0)

PR #145, opened 2025-10-30
fix: Add 'Remove from Continue Watching' option on right-click

SwiftExplorer567:
## Summary
This PR fixes an issue where the 'Remove from Continue Watching' option only appeared when clicking the three-dot menu button, but not when right-clicking directly on an item card.

## Changes
- Added a `contextmenu` event listener to detect right-clicks on Continue Watching items
- Refactored the detection logic into a reusable `checkAndSetContinueWatchingContext` helper function
- Now works consistently for both three-dot menu clicks and right-clicks on cards

## Testing
Tested and confirmed working:
- ✅ Right-click on Continue Watching item → 'Remove from Continue Watching' option appears
- ✅ Click three-dot menu button → 'Remove from Continue Watching' option appears (existing functionality preserved)

## Files Changed
- `Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/events.js`

This provides a more intuitive user experience by allowing users to access the remove option via right-click, which is a common pattern in desktop applications.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/events.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/events.js
@@ -190,17 +190,14 @@
      * Listens for context menu clicks to identify "Continue Watching" items.
      */
     function addContextMenuListener() {
-        document.body.addEventListener('mousedown', (e) => {
-            if (!JE.currentSettings.removeContinueWatchingEnabled) return;
-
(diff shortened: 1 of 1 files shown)


SwiftExplorer567:
I have used Claude Code 4.5, please make sure everything is all good. I have tested on my server.
it was working as intended.

PLEASE review the code, AI is used.


Jellyfin 10.11.1
windows 11
chrome

PR #183, opened 2025-11-15
Fix watchlist following to https://github.com/ranaldsgift/KefinTweaks/commit/216be4502a87730da43da01aabdb9a86b50249f0

66Bunz:


66Bunz:
> Should it be `/scripts/` with a slash at the end since the orig was a slash?

That's true, fixed it

n00bcodr:
Thanks @66Bunz for the PR. Although, I won’t be merging it. The Watchlist feature will be removed from Jellyfin Enhanced in the next update. 

It was always borrowed from [KefinTweaks](https://github.com/ranaldsgift/KefinTweaks), and now that KefinTweaks is stable with watchlist and so many more awesome tweaks, watchlist will be handed back to @ranaldsgift :)

Issue #152, opened 2025-10-31
➕[Feature Request] Make Jellyseerr search's result cards' images clickable


dfvan:
When using the Jellyseerr search, on the result cards, only the title and the request buttons are clickable. Instinctively I click on the image instead of the title to get more details. Would it be possible to extend the behavior from the title also to the card's image please?

n00bcodr:
I did that on purpose, although I don't remember why now! 😅

n00bcodr:
Yeah, the reason I had them the way they are was to make it easier to see the overview on mobile. 

And also, instead of going into an external link, exiting the app or moving out of the webpage whenever someone clicks on the poster by muscle memory, this helps keep the experience within the app. 

That being said I am inclined not to do it, because I like the behavior as is. Sorry about that 😓 


But a hack would be extending the link to the card using a pseudo element

```css

.jellyseerr-card .cardText-first > a[is="emby-linkbutton"] {
    position: static !important;
}

.jellyseerr-card .cardText-first > a[is="emby-linkbutton"]::before {
    content: '';
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;
    height: 50%;
    z-index: 1;
}

.jellyseerr-card .cardFooter {
    position: relative;
    z-index: 2;
}

```


You can play around with the `top` and `height` values

But that would make the top 50% of the card clickable (Green for illustration) 

<img width="561" height="578" alt="Image" src="https://github.com/user-attachments/assets/5f600e7a-0661-41c6-a032-cdefcbb6167a" />

dfvan:
Thank you, that is a great workaround!

akash-258:
@n00bcodr What about this approach :

The current request popup can be upgraded with complete/more metadata including title description etc, and this should be used as the default popup for both, when clicking on the poster and the request button (cases where request button needs a popup) ?

n00bcodr:
@akash-258 If I am being completely honest.

I am not following.

And I think we should take this on your issue itself, because what this current Feature Request has asked and what you are asking might be completely different?

PR #185, opened 2025-11-17
[Fix] Allow Entire Request Cover to Be Clicked

zcraftelite9495:
* Set it up to allow the entire request cover to be clicked to go to IMDB/Jellyseer as opposed to just the title.


zcraftelite9495:
Closed, #152.

Issue #136, opened 2025-10-28
➕[Feature Request] Show filesizes for Entire Shows and Seasons


GithubEnjoye:
**Describe the bug**
Those settings only work with Movies they don't work with TV Shows:
 UI Settings
Show file sizes
Show available audio languages on item detail page

**Expected behavior**
Settings work on movies and shows.

**Jellyfin Environment (please complete the following information):**
 - Server Version: 10.11.0
 - Plugin Version: 10.11.0
 - Jellyfin Installation Method: Linux
 - Device: Microsoft Edge Web Browser PC
 - Theme being used: -

n00bcodr:
How do you mean don't work, what do they show?

GithubEnjoye:
on the show screen I see:
2019 - Present TV-MA 8.2 {Total File Size (icon)} - {Audio Language (icon)} -
instead of:
2019 - Present TV-MA 8.2 {Total File Size (icon)} 1.01 GB {Audio Language (icon)} English

on the movies screen it works

n00bcodr:
You looking at the season or the show page, it wont work there, it should work in individual episodes page.

GithubEnjoye:
Would it be possible to hide it on this screen if there is no way to retrieve and show the sum of all episodes and the language options of the first episode?

n00bcodr:
Right, I thought about this, made some changes and tested.

This might potentially cause problems if implemented. 

For example, Imagine having a show, with 100s of episodes, we'd essentially be making an API call for the size of each episode and waiting for all of them to compute the size.

Although it might not take that long, it is a lot of unnecessary traffic IMO. Will investigate if there are other ways. But right now this is where we are.

Also I will be changing this to a Feature Request, as it is doing what it is intended to do right now.

n00bcodr:
Thanks to @Namo2, this should be now included in version [9.5.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/tree/9.5.0.0)

PR #191, opened 2025-11-20
feat: file size info for shows and seasons

Namo2:
This PR moves the current file size calculation to the Backend.
This allows us to calculate file sizes for shows and seasons.
Other types like Videos/Folders could be added later, for now they will follow the same behaviour as in the old implementation


Namo2:
For #136

PR #201, opened 2025-11-25
Add Russian language support for genre icons

DanteMS:
This adds Russian language support for genre icons.
Translations were taken from TMDB.


n00bcodr:
Thank you!

PR #240, opened 2025-12-26
Improve issue reporting: always search missing episodes & lazy TMDB resolution

4eh5xitv6787h645ebv:
### Auto-search on Issue Report (new feature)
- **Automatic search trigger** - When users report an issue via the "Report Issue" button, optionally triggers an automatic search in Sonarr/Radarr to find a better quality version of the problematic media
- **Sonarr integration** - Supports series, season, and episode-level searches for TV shows
- **Radarr integration** - Supports movie searches
- **Missing episode search** - Triggers search for missing episodes when reporting TV show issues
- **Configuration option** - New `IssueAutoSearchEnabled` setting with UI toggle
- Uses existing Sonarr/Radarr URLs and API keys from *arr Tags Sync settings

### Improvements & Fixes
- **Missing episode search for any season** - Previously only triggered when reporting at series-level. Now triggers for any TV show issue, since users can only report on seasons in Jellyfin but other monitored seasons may have missing episodes
- **Lazy TMDB ID resolution** - Removed upfront TMDB ID check from availability. TMDB ID resolves on button click with fallbacks (parent lookup, search API). Fixes button not appearing or appearing blank when TMDB ID wasn't immediately available
- **Always show "All seasons" option** - Previously only shown with multiple seasons, now always available
- **Improved defaults** - Series-level views default to "All seasons" and "All episodes"
- **Retry logic for button placement** - Up to 5 retry attempts with increasing delay when button container isn't ready

## Test plan
- [x] Enable auto-search in plugin settings
- [x] Report issue on a movie - verify Radarr search triggers
- [x] Report issue on a TV show (any season) - verify Sonarr search and missing episode search trigger
- [x] Test on items without direct TMDB IDs - button should appear and resolve on click
- [x] Verify button no longer appears blank or missing
- [x] Verify button appears correctly on slow-loading pages

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Services/SonarrService.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Services/SonarrService.cs
@@ -241,12 +241,10 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Services
                 {
                     _logger.Info($"Sonarr {commandName} triggered successfully for '{series.Title}'");
 
-                    // For series-level issues (no specific season), also trigger missing episode search
+                    // Always trigger missing episode search for TV shows
                     // This helps find episodes that failed to download across all monitored seasons
-                    if (seasonNumber == 0)
-                    {
-                        await TriggerMissingEpisodeSearchAsync(httpClient, commandUrl, series.Id, series.Title);
-                    }
+                    // (user can only report on seasons that exist in Jellyfin, but other seasons may be missing)
+                    await TriggerMissingEpisodeSearchAsync(httpClient, commandUrl, series.Id, series.Title);
 
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


MicrowavedBunny:
I like the addition of optional auto searching! Does this would it be able to overwrite the original issued version automatically?
Some nice fixes an improvements!

Is the purpose of having "all seasons" always available to say episodes/seasons are missing? for example: having only season 5 of a series and they reporting all seasons to mean seasons 1-4 are missing? I don't know if that is really an issue. Could be purposeful I'm unsure of the point of having it always available. Would it also show the single season that is available for example "Season1" + "all seasons", that would be redundant but if not then all seasons isn't very descript even if there is only 1 season. 

any show with more than a single season already defaults to "All seasons" and "All episodes" but I can see this change making sense if the previous change is implemented.

PR #259, opened 2026-01-06
Feat/watch-progress-as-time-string

Namo2:
Fix: Show watch progress, language tags, file size for seasons

Feat:
Adds the abbility to click on the watch progress to switch between percentages (default, which was already implemented)
and a time based string

image

image

In my opinion this is the cleanest way to implement this on the UI,
if you think this fits @n00bcodr let me know.
We could also add a setting, to change the default display (time string instead of percentages)


n00bcodr:
Will we give an option to select between % or time?

PR #294, opened 2026-01-24
Fix config banner text light/dark

lilyMrt:
Text is swapped

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.html
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.html
@@ -1066,7 +1066,7 @@
                                     </div>
                                     <div style="display: flex; flex-direction: column; gap: 0.5em; height: 100%;">
                                         <label style="font-weight: bold; color: #eee;">Banner Light</label>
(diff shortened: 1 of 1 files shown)


n00bcodr:
Not really though. Banner Light **is** for Dark Mode and Banner Dark **is** for Light mode. What I have is correct?

[_base/_theme.scss#L76](https://github.com/jellyfin/jellyfin-web/blob/6b90debcf78fabe2733de4fc08580880bfbf8ab7/src/themes/_base/_theme.scss#L76)

lilyMrt:
> Not really though. Banner Light **is** for Dark Mode and Banner Dark **is** for Light mode. What I have is correct?
> 
> [_base/_theme.scss#L76](https://github.com/jellyfin/jellyfin-web/blob/6b90debcf78fabe2733de4fc08580880bfbf8ab7/src/themes/_base/_theme.scss#L76)

Just doesn't really make sense to me semantically, why is 'light' banner used for dark mode and vice versa?
Perhaps I also need to update the theme

n00bcodr:
> why is 'light' banner used for dark mode and vice versa?

You want light text on dark backgrounds and vice versa. If you think of it that way, it makes sense. 

Anyway. Going to close this PR.

PR #327, opened 2026-01-30
Keep last calendar view in user settings

Zenrac:
## Summary
This PR allows to keep the user's last selected calendar view.
The selected value is persisted in user `settings.json` configuration and applied consistently on the calendar page.

### Changes
- Added support for `CalendarDefaultViewMode`
  - If value is `week`, `month`, or `agenda` → use directly
  - If value is `auto` or missing → agenda on mobile, month on desktop

### Tests
- [x] Going to another view updates the configuration
- [x] Configuration is kept after refresh
- [x] Default view is selected according to the current configuration


n00bcodr:
This can be a  user-level parameter

We can save the view in settings.json when the view changes and load that, lmk what you think.

Zenrac:
@n00bcodr was this what you asked?

n00bcodr:
Yes!
I think this is better, instead of admin setting.

n00bcodr:

PR #328, opened 2026-01-30
fix: Jellyseerr proxy URL fallback on HTTP errors

### LINKED ISSUE #321: ➕[Feature Request] Spinner for jellyseer search results
**Is your feature request related to a problem? Please describe.**
My jellyfin-enhanced is quite slow for showing jellyseerr results in search (>30 secs... is that normal?). There is no visual feedback to indicate that it's loading so I don't know whether it has crashed or there are no results

**Describe the solution you'd like**
I think it would be best if the jellyseerr section was always present and show a spinner while loading, instead of just popping when the search is completed

**Additional context**
If there's interest in this I may be able to vibe a PR

4eh5xitv6787h645ebv:
## Summary

- Fixes `ProxyJellyseerrRequest()` to continue trying the next configured Jellyseerr URL when the current one returns a non-2xx HTTP response, instead of returning the error immediately
- Aligns proxy fallback behavior with `GetJellyseerrStatus()`, which already falls through correctly on HTTP errors
- If all URLs fail, returns the last error's status code instead of a generic 500

## Problem

When multiple Jellyseerr URLs are configured and the first URL is reachable but returns an HTTP error (e.g. 404 from a wrong server), `ProxyJellyseerrRequest()` immediately returns that error to the client without trying the remaining URLs. This breaks search, requests, and all other Jellyseerr proxy endpoints.

The fallback only worked for **connection failures** (DNS errors, timeouts) because those throw exceptions caught by the `catch` block. HTTP error responses took a different code path that returned immediately.

| Failure type | Before | After |
|---|---|---|
| Connection exception (DNS, timeout) | Falls through to next URL | Falls through to next URL |
| HTTP error (404, 500, etc.) | **Returns immediately** | Falls through to next URL |

## Implementation Notes

This PR was developed with AI assistance (Claude). All changes have been reviewed, tested, and the implementation is understood.

## Test plan

- [x] Configured two bad URLs (`https://google.com`, `https://googasdggle.com`) + one valid URL (`http://jellyseerr:5055`)
- [x] Verified server logs show all three URLs attempted in sequence
- [x] Verified search results render successfully after fallback to the working URL
- [x] Verified `GetJellyseerrStatus()` behavior is unchanged (already correct)
- [x] No console errors
- [x] Tested on Jellyfin 10.10

Closes #321

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
@@ -175,6 +175,9 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
                 return BadRequest(new { message = "Jellyfin User ID was not provided in the request." });
             }
 
+            int lastStatusCode = 500;
+            string lastErrorContent = "Could not connect to any configured Jellyseerr instance.";
+
             foreach (var url in urls)
             {
                 var trimmedUrl = url.Trim();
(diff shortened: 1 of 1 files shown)


n00bcodr:
Thank you!

n00bcodr:

PR #332, opened 2026-01-30
Use new item by provider id endpoint to navigate in calendar

Zenrac:
## Summary

Use `/JellyfinEnhanced/items/by-providers` to find Jellyfin Item rather than searching it by title (but keep this as fallback).

## Tests
- [x] Navigation works and API call is displayed in console network


n00bcodr:

Issue #360, opened 2026-02-07
➕[Feature Request] Further integration with home screen sections


AdamWHY2K:
**Is your feature request related to a problem? Please describe.**
No, there are a couple of open feature requests on Paradox's site that seem like they would be better suited here:
[Show ratings on posters on the home page](https://features.iamparadox.dev/posts/66/show-ratings-on-posters-on-the-home-page)
[Home Screen Sections "Discover" cards - Jellyfin Enhanced "More Info" modal](https://features.iamparadox.dev/posts/68/home-screen-sections-discover-cards-jellyfin-enhanced-more-info-modal)

**Describe the solution you'd like**
Features linked above implemented 👉👈 

**Additional context**
Completely understand if you disagree, happy to close and implement on HSS if you'd prefer 😊

n00bcodr:
Ahh, I will leave homescreen sections to our lord @IAmParadox27

[Show ratings on posters on the home page](https://features.iamparadox.dev/posts/66/show-ratings-on-posters-on-the-home-page) is easily doable.

[Home Screen Sections "Discover" cards - Jellyfin Enhanced "More Info" modal](https://features.iamparadox.dev/posts/68/home-screen-sections-discover-cards-jellyfin-enhanced-more-info-modal) I either can add an option in my plugin or in HSS plugin to open the more-info-modal for HSS cards

IAmParadox27:
@AdamWHY2K I'm happy to rework the discover cards in HSS and behave this way (potentially working with @n00bcodr to create an interface between the plugins) if JFE exists

n00bcodr:
https://github.com/user-attachments/assets/40f5e0f2-2009-4241-be37-a3c7766a2549

Couldn't send it on Discord due to file size!

@IAmParadox27 

These changes should be fine with or without Jellyfin Enhanced, I will add some gating and warnings in JE on when this will work.

https://github.com/IAmParadox27/jellyfin-plugin-home-sections/pull/202

n00bcodr:
Done, one change is merged another one has a PR. Going to close this :)

n00bcodr:
In version [11.0.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/releases/tag/11.0.0.0)

Will kick in once HSS is released.

PR #372, opened 2026-02-09
Add HSS discovery handler for Jellyseerr modal

n00bcodr:
Introduce a new hss-discovery-handler.js that intercepts clicks on Home Screen Sections (discover-card) and opens the Jellyseerr more-info modal (unless the request button is clicked).

For #360

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/jellyseerr/hss-discovery-handler.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/jellyseerr/hss-discovery-handler.js
@@ -1,0 +1,47 @@
+/**
+ * Home Screen Sections (HSS) Discovery Handler
+ * Intercepts discover card clicks and opens the Jellyseerr more-info modal
+ * instead of navigating to the Jellyseerr website
+ */
+
+(function(JE) {
+    'use strict';
+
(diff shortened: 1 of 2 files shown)


n00bcodr:
@IAmParadox27 here is the handler to intercept the request. Also only works if https://github.com/IAmParadox27/jellyfin-plugin-home-sections/pull/202 is available, else just falls back to Jellyseerr URL

Still contemplating if should add a plugin config option to enable this or not.

Thoughts @AdamWHY2K?

AdamWHY2K:
Yeah, I think more config options are always good. I think this one can be on by default though.

n00bcodr:
Yeah, this handles fallback itself, only opens the more-info-modal if it is available (enabled)

I was just thinking a config option would cause confusion to the end user.

AdamWHY2K:
Ah yeah, if this checks the existing 'open in more info modal' setting (sorry I haven't had time to look at the code yet) then I don't think we need another one, but I do think that setting should be moved out of the search section.

n00bcodr:
> Ah yeah, if this checks the existing 'open in more info modal' setting (sorry I haven't had time to look at the code yet) then I don't think we need another one, but I do think that setting should be moved out of the search section.

Moved now

image

PR #338, opened 2026-01-31
Rework Calendar UI

Zenrac:
## Summary
This PR reworks the calendar UI and adds richer ARR metadata to improve presentation and navigation.
It introduces per-user display mode defaults and new locale strings for day labels.

### Changes
- Calendar UI overhaul (views, layout, cards, agenda)
- Removes old "search" fallback to get item by provider id
- Reverts UI UTC dates to local dates (parameters are converted when sending to API)
- Adds `CalendarDisplayMode` (screen settings) that displays calendar by either List, Backdrop, Card
- Adds "Day" view
- Adds `BackdropUrl` to calendar items and populated posters/backdrops from Sonarr/Radarr

### Tests
- [x] Plugin Calendar Page works
- [x] Normal Calendar Page works

PR #388, opened 2026-02-13
fix: Calendar card click navigates to episode instead of series

4eh5xitv6787h645ebv:
## Summary

Clicking a calendar episode card currently navigates to the **series** detail page instead of the **specific episode**. This is because `handleEventClick` passes `{ preferSeries: true }` to `navigateToJellyfinItem`.

This changes it to `{ preferSeries: false }` so the click resolves to `event.itemEpisodeId` (the episode) first, falling back to series providers only if no episode match is found. Movies are unaffected.

**Question:** Was the original `preferSeries: true` behavior intentional — i.e., card click = series page, play button = episode? Or should clicking an episode entry in the calendar take you directly to that episode?

### How it broke

| Commit | What happened |
|--------|--------------|
| `e3f662c` (PR #332) | Added provider-based navigation — no `preferSeries` concept, worked fine |
| `9863ec9` (PR #338) | Rework Calendar UI — introduced `preferSeries`, set card click to `true` |
| `c6433ca` | Added `itemEpisodeId` support — play button (`preferSeries: false`) navigates to episode correctly, but general card click stayed `preferSeries: true` |

### One-line change

```diff
- navigateToJellyfinItem(event, { preferSeries: true });
+ navigateToJellyfinItem(event, { preferSeries: false });
```

## Test plan
- [x] Click an episode card in calendar → should navigate to episode detail page
- [x] Click play button on episode → should navigate to episode detail page  
- [x] Click a movie card in calendar → should navigate to movie detail page (unchanged)
- [x] Verify series-level navigation still works via other UI paths

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar-page.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar-page.js
@@ -2705,7 +2705,7 @@
 
     e.preventDefault();
     e.stopPropagation();
-    navigateToJellyfinItem(event, { preferSeries: true });
+    navigateToJellyfinItem(event, { preferSeries: false });
   }
 
   /**


4eh5xitv6787h645ebv:
or should this be a new setting for the user to chose? so they could decide it it should just go to the series for to the exact episode?

4eh5xitv6787h645ebv:
Moving this to an issue instead (#389) to discuss whether this should be a user setting.

Issue #413, opened 2026-02-24
🐛[BUG] Subtitle setting Disable custom styling not correctly respecting original settings


NoudH:
**Describe the bug**
When disabling custom subtitle styling, the original subtitle settings are not respected.
Subtitle styles are written to `#htmlvideoplayer-cuestyle`. When the custom subtitle styling is disabled, it still deletes the styling in this stylesheet.
I believe the issue is related to https://github.com/n00bcodr/Jellyfin-Enhanced/blob/18f94f362845d6c6cc92c5448ac795d3c35411b2/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/subtitles.js#L112

**Expected behavior**
When disabling custom subtitle styling, it should use and respect the default Jellyfin settings.

**Screenshots**
What happens when you disable custom subtitle styling:

<img width="1027" height="293" alt="Image" src="https://github.com/user-attachments/assets/8d56248c-4784-4579-aade-927a46674515" />

What it should be (I've set the font size setting in Jellyfin to extra large for emphasis):

<img width="1155" height="275" alt="Image" src="https://github.com/user-attachments/assets/516e41fc-afe3-4cb4-b2d7-94ecb615e92b" />

As you can see, the font-family and font-size are different.

**Jellyfin Environment (please complete the following information):**
 - Server Version: 10.11.6
 - Plugin Version: 11.0.0.0
 - Jellyfin Installation Method: Docker (linuxserver/jellyfin)
 - Device: Jellyfin Web on Chrome
 - External Access Method: N/A
 - Theme being used: ElegantFin

**Logs**

N/A

**Additional context**
Perhaps a better implementation would be to apply a more specific stylesheet when applying custom subtitle styling, instead of deleting the original styles. I.e. creating a new style element with the selector `video.htmlvideoplayer::cue` and adding the custom styling to it. Then when the option is disabled, just remove the custom style element.

NoudH:
I've created a pull request that fixes this issue: #414

NoudH:
Seeing as the PR has been merged I'll close the issue.

@n00bcodr any ETA when the new update with this fix will be released?

n00bcodr:
Just busy with work!

I will see if I can get sometime to release it during this week, if not Friday :(

I just want to get a couple more bugfixes sorted before releasing a version. Apologies!

PR #414, opened 2026-02-24
fix(subtitles): disable custom subtitles now respects default subtitle settings

NoudH:
Fixes the issue mentioned in https://github.com/n00bcodr/Jellyfin-Enhanced/issues/413


n00bcodr:
Thank you!

PR #418, opened 2026-02-26
fix: prevent duplicate Jellyseerr requests from auto-season-request

### LINKED ISSUE #398: 🐛[BUG] Automatic Requests requests content again
**Describe the bug**
I watched season 3 (final season btw) of Vox Machina last week. And apparently it requests season 4 (which isnt even announced but I don't mind that) multiple times during that time (see screenshots)...

**Expected behavior**
Ideally it would only request each item only once.
To futureproof for when the seerr project inevitably implements multi user requests the plugin should check if the current user has already requests the thing it is going to request before requesting it

**Screenshots**
Image

 - Server Version: 10.11.6
 - Plugin Version: 11.0.0.0
 - Jellyfin Installation Method: truenas scale native docker

4eh5xitv6787h645ebv:
## Summary

Fixes #398 — Auto-Season-Request creates duplicate Jellyseerr requests every time an episode is played/stopped.

**Three root causes identified and fixed:**

1. **Wrong JSON path** (`AutoSeasonRequestService.cs`): The code checked `root.TryGetProperty("requests", ...)` but the Jellyseerr `/api/v1/tv/{id}` endpoint returns `requests` under `mediaInfo`, never at root level. This meant the existing-request check **always returned false**, so the plugin thought no requests existed even when they did.

2. **No dedup on `PlaybackStopped`** (`AutoSeasonRequestMonitor.cs`): `OnPlaybackProgress` had a `_checkedSessions` dedup dictionary but `OnPlaybackStopped` did not. Multiple stop events for the same episode (which Jellyfin fires 2-4 times per playback) each triggered a separate request.

3. **No in-memory request cache** (`AutoSeasonRequestService.cs`): There was no cache to prevent redundant requests between the progress and stop handlers within the same playback session. Also, the existing movie cache in `AutoMovieRequestService.cs` had no thread safety.

## Changes

- **`AutoSeasonRequestService.cs`**: Fixed JSON path to `mediaInfo.requests`. Added thread-safe `_requestedSeasons` cache with sentinel pattern (keyed globally by `tmdbId_seasonNumber`, not per-user). Sentinel is written before async work so concurrent callers see it immediately; removed on failure to allow retries.

- **`AutoSeasonRequestMonitor.cs`**: Added `_checkedSessions` dedup to `OnPlaybackStopped` with `stopped_` prefix key, matching the existing `OnPlaybackProgress` pattern.

- **`AutoMovieRequestService.cs`**: Added `lock (_movieCacheLock)` around all `_requestedMovies` access. Applied the same sentinel pattern. Added expired-entry cleanup sweep across all users (previously only cleaned accessed user's entries).

## How to reproduce

1. Have a TV show where the current season has enough episodes watched to meet the threshold (e.g., threshold=2, play episode 9 of a 10-episode season)
2. Simulate a single playback: start → progress → stop → stop (Jellyfin fires multiple stop events)
3. **Before fix**: Plugin logs show `Jellyseerr reports no requests` (wrong path) and creates 3 duplicate `✓ Requested` entries. Jellyseerr receives 3 separate POST requests.
4. **After fix**: Plugin logs show `Jellyseerr reports N request(s)` (correct path). Only 1 request is created; subsequent events are blocked by cache (`Already requested S2 (cached)`) and session dedup (`PlaybackStopped already processed`).

### Reproduction logs (buggy code)
```
[13:36:30] ✓ Requested 'The Expanse' S2  ← OnPlaybackProgress
[13:36:34] ✓ Requested 'The Expanse' S2  ← OnPlaybackStopped #1 (no dedup)
[13:36:38] ✓ Requested 'The Expanse' S2  ← OnPlaybackStopped #2 (no dedup)
Jellyseerr: 3 duplicate S2 requests (IDs 204, 205, 206)
```

### Fixed code logs (same test)
```
[13:46:30] ✓ Requested 'The Expanse' S2  ← OnPlaybackProgress (legitimate)
[13:46:34] Already requested S2 (cached)  ← blocked by sentinel cache
[13:46:37] PlaybackStopped already processed ← blocked by session dedup
Jellyseerr: 1 S2 request
```

## Why it happened

The Jellyseerr API structure was the core issue. The `/api/v1/tv/{id}` endpoint returns:
```json
{
  "id": 63639,
  "name": "The Expanse",
  "seasons": [...],
  "mediaInfo": {
    "requests": [...]  // ← requests are HERE
  }
}
```

The old code used `root.TryGetProperty("requests", ...)` which always returned `false` since `requests` doesn't exist at root level. This made the plugin think every show had zero existing requests, causing it to re-request on every playback event.

Combined with no dedup on `PlaybackStopped` and no in-memory cache, a single episode playback could generate 3-4 duplicate requests (1 from progress + 2-3 from stop events).

## Test plan

- [x] Verified `root.requests` does NOT exist in Jellyseerr API response (confirmed via live API call)
- [x] Reproduced 3 duplicate requests from single playback with unmodified code
- [x] Confirmed Jellyseerr created actual duplicate entries (Request IDs 204, 205, 206 for The Expanse S2)
- [x] Fixed code: only 1 request created despite 4 playback events (1 progress + 1 start + 2 stops)
- [x] After container restart (cache cleared), API-only check (`mediaInfo.requests`) still prevents duplicates
- [x] Sentinel pattern prevents race conditions between concurrent handlers
- [x] Cache expires after 1 hour, allowing legitimate future requests
- [x] Plugin loads cleanly with no errors
- [x] Compatible with Jellyfin 10.11.6

**AI-assisted**: This PR was developed with Claude Code (Opus 4.6) assistance. All code was reviewed for correctness and tested against a live Jellyfin/Jellyseerr environment.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Services/AutoMovieRequestService.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Services/AutoMovieRequestService.cs
@@ -93,28 +93,34 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Services
             }
 
             // Check if we've already requested this movie (in-memory cache with 1-hour expiry)
+            // Uses a sentinel pattern: write the entry before async work so concurrent
+            // callers see it immediately, then remove on failure to allow retries.
             var requestKey = $"{user.Id}_{nextMovieInfo.TmdbId}";
             lock (_movieCacheLock)
             {
+                // Clean up expired entries across all users
+                foreach (var cachedUserId in _requestedMovies.Keys.ToList())
+                {
+                    var expired = _requestedMovies[cachedUserId]
+                        .Where(kvp => (DateTime.Now - kvp.Value).TotalHours >= 1)
+                        .Select(kvp => kvp.Key).ToList();
+                    foreach (var key in expired) _requestedMovies[cachedUserId].Remove(key);
+                    if (_requestedMovies[cachedUserId].Count == 0) _requestedMovies.Remove(cachedUserId);
+                }
+
                 if (!_requestedMovies.ContainsKey(user.Id.ToString()))
                 {
                     _requestedMovies[user.Id.ToString()] = new Dictionary<string, DateTime>();
                 }
 
-                // Check if cached and not expired (1 hour)
-                if (_requestedMovies[user.Id.ToString()].TryGetValue(requestKey, out var cachedTime))
+                if (_requestedMovies[user.Id.ToString()].ContainsKey(requestKey))
                 {
-                    if ((DateTime.Now - cachedTime).TotalHours < 1)
-                    {
-                        _logger.Debug($"[Auto-Movie-Request] Already requested '{nextMovieInfo.Title}' (cached)");
-                        return;
-                    }
-                    else
-                    {
-                        // Expired, remove from cache
-                        _requestedMovies[user.Id.ToString()].Remove(requestKey);
-                    }
+                    _logger.Debug($"[Auto-Movie-Request] Already requested '{nextMovieInfo.Title}' (cached)");
+                    return;
                 }
+
+                // Reserve the slot so concurrent callers see it immediately
+                _requestedMovies[user.Id.ToString()][requestKey] = DateTime.Now;
             }
 
             // Request the movie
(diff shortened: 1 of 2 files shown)


n00bcodr:
As always!

Thank you! :)

PR #464, opened 2026-03-17
Update languagetags.js

Kisnov:
I changed the website where it takes language flag images from, so it has more language flags than just languages with a country, and added 3 new language flags. The changes are minimal I hope this is enough for it to work.


n00bcodr:
Thank you! :)

PR #469, opened 2026-03-18
fix: prevent plugin from overwriting user's language on every page load

### LINKED ISSUE #441: 🐛[BUG] Language Settings in JE doesn't respect override user sets
**Describe the bug**
When using vanilla JellyFin using:
```bash
docker run --rm -it \
  --name jellyfin-temp \
  -p 8097:8096/tcp \
  -e UMASK=002 \
  --device /dev/dri:/dev/dri \
  docker.io/jellyfin/jellyfin:10.11
```

I can set the language on a user level, and no matter how many times i refresh or even use CRTL F5, it will remember my language option.

When i install Jellyfin-Enhanced and not configure anything, restart and navigate as the user back, i do notice the language is set to automatic (intended). I then set it to English for example, i save it. And after a refresh it's in Engilsh!
But if i refresh again, it's back to automatic! This is tested locally and directly to the container.

**Expected behavior**
`Sets the default language for all users. Users can still override this in their own settings.`
I expect just like Jellyfin vanilla does, that the language as a user i set, is remembered.

**Screenshots**

Image

Image

Image

So far expected.

Image

This is good still

Image

This is odd, this reverted back again?

Image

Language is set to dutch again after a refresh.

 - Server Version: 10.11.6
 - Plugin Version: 11.2.0.0
 - Device: Jellyfin Web on Chrome
 - External Access Method: Direct, nothing in between and locally
 - Theme being used: Vanilla/none

**Logs**

Please provide server logs and browser console logs as applicable (make sure sensitive information is redacted)

[Full-jellyfin-logging.txt](https://github.com/user-attachments/files/25840565/Full-jellyfin-logging.txt)
^hope this helps, but in here and in the browser console i couldn't find anything odd/related myself. But i hope it's as easily reproducable on your side aswell!

**Additional context**

4eh5xitv6787h645ebv:
## Description

Fixes #441

The plugin was unconditionally writing to the `{userId}-language` localStorage key on every page load, comparing it against `JE.currentSettings.displayLanguage` and overwriting if different. When `displayLanguage` is empty (the default "Auto"), this wrote `''` to localStorage, erasing whatever language the user had set via Jellyfin's native settings UI. The user's language would revert to "Auto" after 2 refreshes.

## Implementation

Instead of syncing localStorage on every page load, the plugin now only seeds the admin's default language when the user has no language set yet (`localStorage.getItem(languageKey) === null`). Once any language is set by the plugin on first visit, by the user via Jellyfin's native settings, or via the JE panel the plugin never overwrites it.

No additional localStorage keys are introduced. The fix is a single-file change in `plugin.js`.

## Implementation Notes

This PR was developed with AI assistance (Claude). The change has been reviewed, tested, and verified on a live Jellyfin 10.11 instance.

## Testing

- [x] Verified no console errors
- [x] Reproduced the bug with unmodified main branch (language reverts after 2 refreshes)
- [x] Confirmed fix resolves the issue (language persists across multiple refreshes)
- [x] Tested with admin and non-admin users
- [x] Doesn't break existing functionality (JE panel language dropdown still works)
- [x] Build passes with 0 warnings, 0 errors

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/plugin.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/plugin.js
@@ -554,20 +554,20 @@
 
             if (userId) {
                 const languageKey = `${userId}-language`;
-                const storedLanguage = localStorage.getItem(languageKey) || '';
-                const desiredLanguage = (JE.currentSettings?.displayLanguage || '').trim();
-                const normalizeLangCode = (code) => {
-                    if (!code) return '';
-                    const parts = code.split('-');
-                    if (parts.length === 1) return parts[0].toLowerCase();
-                    if (parts.length === 2) return `${parts[0].toLowerCase()}-${parts[1].toUpperCase()}`;
-                    return code;
-                };
-                // Use the language code as-is, no special mapping
-                const desiredStorageLanguage = desiredLanguage ? normalizeLangCode(desiredLanguage) : '';
-
-                if (storedLanguage !== desiredStorageLanguage) {
-                    localStorage.setItem(languageKey, desiredStorageLanguage);
+                // Only seed the admin's default language if the user has no language set yet.
+                // This prevents overwriting the user's own language choice on every page load.
+                if (localStorage.getItem(languageKey) === null) {
+                    const desiredLanguage = (JE.currentSettings?.displayLanguage || '').trim();
+                    if (desiredLanguage) {
+                        const normalizeLangCode = (code) => {
+                            if (!code) return '';
+                            const parts = code.split('-');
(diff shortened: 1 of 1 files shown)


n00bcodr:
But now the problem would be, Once the end user changes the language, the "Apply above settings for all users" would be essentially useless for languages?

4eh5xitv6787h645ebv:
Yes thats true.. i didn't think you would want the "Apply above settings for all users" be applied for languages as well. It would still change it for users that haven't set their language yet though. Or you wanting to to be a force override?

n00bcodr:
Ahh. you know what. 

I think this is a safe tradeoff for the issue. Going ahead with merge :)

PR #490, opened 2026-03-24
Catalan Translation

Kisnov:
I added the json for Catalan language.


Kisnov:
This error is because i deleted {count} because  the plural is not like english season(s) i would have to change the word, not just add another word after it. AI proposed this (but it adds another line, and idk if thats possible: "jellyseerr_modal_toast_request_success_one": "{{icon:tv}} S'ha sol·licitat {count} temporada de \"{title}\"",
"jellyseerr_modal_toast_request_success_other": "{{icon:tv}} S'han sol·licitat {count} temporades de \"{title}\""

Anyway how i wrote it is the best way to put it, but would be losing the count of seasons, if its imposible it can go with: 

"jellyseerr_modal_toast_request_success": "{{icon:tv}} Sol·licitud enviada: {count} temporades de \"{title}\""

Season = Temporada
Seasons = Temporades

Kisnov:
well i will change it so it doen´t give errors, just for some letters... it's not even a big mistake

n00bcodr:
Thank you! :)

PR #506, opened 2026-03-29
Changed 2 translations that werent working.

Kisnov:
One was identical but in minus, so i put it in mayus, i thought it was case insensitive.


n00bcodr:

PR #511, opened 2026-03-30
Reduce size of quality tags or show only inside details poster

Kisnov:
*This is just an example, I'll open the coresponding enhancement issue*

The quality tags occupy to much space on the outside poster, and dolby digital+ 2.0 is the worst case of it.

I proposed 2 changes: changing dolby digital to DD, or another change, that I've no idea at all if it works, it was made with AI, and maybe someone wants it like this so should add an additional checkbox on the config to select both or only inside details page. I like the tags, but the top of the poster is where normally goes the faces or title of series, so I find some change would be nice, maybe not what i proposed.. but just to give some ideas.


Kisnov:
You can cancel my PR, was just done as an example, and maybe to help locate where is the place I think need to be changed, in my unexistent knowledge of the matter.

n00bcodr:
All of this can be achieved using CSS - https://n00bcodr.github.io/Jellyfin-Enhanced/advanced/css-customization/?h=css#quality-tags

PR #517, opened 2026-03-31
fix: move non-sensitive derived values from private-config to public-config

### LINKED ISSUE #514: 🐛[BUG]The reviews sections only appear when logged in as an admin user
Hi! 
The reviews sections only appear when logged in as an admin user.
If I log in with a normal user account, those sections don’t load at all.

4eh5xitv6787h645ebv:
## Description

Fixes a bug where non-admin users lost access to TMDB-dependent features (Reviews, Elsewhere) and Seerr deep links because several non-sensitive derived values were only returned from the `private-config` endpoint, which returns an empty `{}` object for non-admin users.

**Root cause:** When the private/public config split was originally introduced, some derived (non-sensitive) values were placed in `private-config` alongside the truly sensitive ones (API keys, internal service URLs). Since `private-config` returns `{}` for non-admin users, these values were `undefined` on the client, causing feature-enabled checks and deep link generation to silently fail.

### Values moved to `public-config`

| Value | Why it is safe | What broke for non-admin users |
|-------|---------------|-------------------------------|
| `TmdbEnabled` | Boolean derived from whether a TMDB API key is configured -- does not expose the key itself | Reviews, Elsewhere, and TMDB features were disabled (`reviews.js`, `elsewhere.js`, `jellyseerr/ui.js` all bail early) |
| `JellyseerrBaseUrl` | First configured Seerr URL -- no API key, just the base URL for deep links | "Open in Seerr" links on search results returned `null`, falling back to `href="#"` (does nothing) |
| `JellyseerrUrlMappings` | Jellyfin-to-Seerr URL pairs for multi-domain setups -- no secrets | URL mapping resolution failed, so users on alternative domains could not get correct Seerr links |

### Values correctly remaining in `private-config`

`SonarrUrl`, `RadarrUrl`, `BazarrUrl`, and their URL mappings stay in `private-config` because the Arr Links feature is explicitly admin-only (`arr-links.js` checks `isAdmin` and bails for non-admin users at line 53).

No frontend changes needed -- the client already reads these from `JE.pluginConfig` which merges both endpoints.

## Implementation Notes

## Testing

- [x] Verified `public-config` returns `TmdbEnabled`, `JellyseerrBaseUrl`, and `JellyseerrUrlMappings` for all users (admin and non-admin)
- [x] Verified `private-config` no longer includes these values (avoids duplication)
- [x] Verified TMDB review proxy endpoint works for non-admin users
- [x] Verified Arr Links URLs remain in `private-config` (admin-only feature)
- [x] No console errors
- [x] Does not break existing functionality (admin users unaffected)

Closes #514

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
@@ -1759,31 +1759,15 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
                 return new JsonResult(new { });
             }
 
-            // Determine the first configured Seerr URL (if any) for client-side deep links
-            string jellyseerrBaseUrl = string.Empty;
-            try
-            {
-                if (!string.IsNullOrWhiteSpace(config.JellyseerrUrls))
-                {
-                    jellyseerrBaseUrl = config.JellyseerrUrls
-                        .Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries)
-                        .Select(u => u.Trim())
-                        .FirstOrDefault() ?? string.Empty;
-                }
-            }
-            catch { /* ignore */ }
-
             return new JsonResult(new
             {
-                // For Arr Links
+                // For Arr Links (admin-only feature)
                 config.SonarrUrl,
                 config.RadarrUrl,
                 config.BazarrUrl,
                 config.SonarrUrlMappings,
                 config.RadarrUrlMappings,
                 config.BazarrUrlMappings,
-                JellyseerrBaseUrl = jellyseerrBaseUrl,
-                config.JellyseerrUrlMappings
             });
         }
         [HttpGet("public-config")]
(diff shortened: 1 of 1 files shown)


4eh5xitv6787h645ebv:
Shall see if any other setting are affected by this type of bug.

4eh5xitv6787h645ebv:
Audited all values in `GetPrivateConfig()` for the same bug pattern. Found two more affected:

- **`JellyseerrBaseUrl`** -- Seerr deep links broken for non-admin users (search result title links fall back to `href="#"`)
- **`JellyseerrUrlMappings`** -- multi-domain Seerr URL resolution broken for non-admin users

Both are non-sensitive (no API keys, just frontend URLs). Moved them to `GetPublicConfig()` alongside `TmdbEnabled`.

The remaining `private-config` values (`SonarrUrl`, `RadarrUrl`, `BazarrUrl` + their mappings) are correctly admin-only since `arr-links.js` explicitly gates on `isAdmin`.

n00bcodr:
Thank you Jake! :)

n00bcodr:

PR #519, opened 2026-03-31
fix: include primary language in translation fallback chain

### LINKED ISSUE #513: 🐛[BUG] Language does not adapt to the settings.
**Describe the bug**
I have noticed that the language can be changed in the webapp, but in the official Android app the language for the two site menu items (calendar and request) is not being adjusted. I have also already cleared the cache under the Android settings which unfortunately didn't help either. What can I provide to help look into what might be causing this?

**Expected behavior**
Change language as aspected.

**Screenshots**

 - Server Version: 10.11.06
 - Plugin Version: 11.04.0.0

4eh5xitv6787h645ebv:
## Description

Fixes a regression in `buildLanguageChain()` where the primary language code was not included in the fallback chain for simple (non-hyphenated) codes like `de`, `fr`, `es`, `it`, etc. This caused all non-English simple language codes to resolve to `["en"]` only, so translations were never loaded for those languages.

The bug was introduced in PR # 466 (zh-CN support). That PR correctly removed the special-case `zh` handling, but also removed the general-purpose line that added the primary language to the chain:

```js
if (normalizedLang) {
    langCodes.push(normalizedLang);
}
```

This line was originally present when the translations module was first created and was essential for all simple language codes to work.

### Before (broken)
| Input | Chain |
|-------|-------|
| `de` | `["en"]` |
| `fr` | `["en"]` |
| `de-DE` | `["de", "en"]` |

### After (fixed)
| Input | Chain |
|-------|-------|
| `de` | `["de", "en"]` |
| `fr` | `["fr", "en"]` |
| `de-DE` | `["de-DE", "de", "en"]` |

Also adds a fallback to `document.documentElement.lang` when the `{userId}-language` localStorage key does not exist. This covers the Android app and other clients where Jellyfin has already resolved the user's preferred language from server-side settings but the per-client localStorage key has not been written yet.

## Testing
- [x] Verified `JE.t("calendar_title")` returns `"Kalender"` (German) after setting language to `de`
- [x] Verified `buildLanguageChain` produces correct chains for `de`, `fr`, `en`, `de-DE`, `en-GB`, `pt-BR`, `zh-CN`
- [x] No console errors

Closes #513

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/translations.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/translations.js
@@ -17,6 +17,10 @@
         const normalizedLang = normalizeLangCode(primaryLang);
         const langCodes = [];
 
+        if (normalizedLang) {
+            langCodes.push(normalizedLang);
+        }
+
         if (normalizedLang && normalizedLang.includes('-')) {
             const baseLang = normalizedLang.split('-')[0];
             if (!langCodes.includes(baseLang)) {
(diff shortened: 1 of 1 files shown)


n00bcodr:

PR #529, opened 2026-04-04
fix: warn admin when jellyfin-web version is behind server

### LINKED ISSUE #528: 🐛 [BUG] Jellyseerr search results not displayed when Meilisearch is enabled - wrong input selector + searchResults hidden
**Describe the bug**
Jellyseerr search results never appear in the search page despite the integration being fully 
operational (API working, userFound=true, active=true). The issue persists after upgrading to 
11.4.0.0 which includes the fix from #473.

**Root cause identified (two distinct bugs):**

**Bug 1 - Wrong input selector**
jellyseerr.js uses `#searchPage #searchTextInput` to detect search input events, but the actual 
element in Jellyfin 10.11.x is:
`<input class="searchfields-txtSearch emby-input" is="emby-input" id="embyinput1">`

The selector `#searchTextInput` returns null, so no input events are ever captured.

Confirmed via console:
`document.querySelector('#searchPage #searchTextInput')` → null
`document.querySelector('#searchPage .searchfields-txtSearch')` → <input> ✅

**Bug 2 - .searchResults container stays hidden**
Even when renderJellyseerrResults() is called manually, results are injected into 
.jellyseerr-section but the parent container .searchResults has class `hide` and the plugin 
never removes it. Results exist in the DOM but are invisible.

Confirmed via console:
`document.querySelector('.searchResults').classList.remove('hide')` → results appear immediately ✅

**Note:** This bug may be specific to installations using Meilisearch. The contributor who 
wrote the fix for #473 mentioned testing with Meilisearch did not reproduce the issue — 
Meilisearch likely changes how .searchResults visibility is managed.

Jellyfin Environment
- Server Version: 10.11.7.0 (also tested on 10.11.x)
- Plugin Version: 11.4.0.0
- Jellyfin Installation Method: Native Linux (Debian 12, apt repository)
- Device: Firefox on Windows
- External Access Method: Nginx reverse proxy + Cloudflare
- Search engine: Meilisearch enabled

Workaround
Manually call in browser console after searching:
document.querySelector('.searchResults').classList.remove('hide')
Results appear immediately. The jellyseerrAPI.search() function works correctly.

4eh5xitv6787h645ebv:
## Summary

Adds a version mismatch warning banner on the admin config page when the `jellyfin-web` version is behind the server version. This was the root cause of #528 -- the user had `jellyfin-web` stuck on `20220818.18-unstable` (a 2022 build) while `jellyfin-server` was on `10.11.7`.

Rather than patching individual selectors that break on older web clients, this surfaces the real problem to the admin so they can fix it at the source.

Would you want something like this in the plugin?
Screenshot_20260404_133331

### How it works
- Compares `ApiClient._appVersion` (web) vs `ApiClient._serverVersion` (server) on config page load
- Only warns when web client major.minor is **behind** the server (web ahead of server does not trigger)
- Same minor with different patch (e.g. 10.11.3 vs 10.11.7) does not trigger
- Handles non-semver version strings (e.g. date-based unstable builds) gracefully

### AI Assistance Disclosure

### Testing
- [x] Deployed and tested on Jellyfin 10.11.7 dev instance
- [x] Tested: matching versions (no warning)
- [x] Tested: web behind server, e.g. 10.8.13 vs 10.11.7 (warning shown)
- [x] Tested: web ahead of server (no warning)
- [x] Tested: same minor, different patch (no warning)
- [x] Tested: non-semver version string `20220818.18-unstable` (warning shown)
- [x] Tested on Chromium and Firefox
- [x] No console errors
- [x] Does not break existing functionality

Closes #528

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.html
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.html
@@ -2494,14 +2494,16 @@
 
                 var webParts = webVersion.split('.').map(Number);
                 var serverParts = serverVersion.split('.').map(Number);
+                var validSemver = !isNaN(webParts[0]) && !isNaN(webParts[1]) && !isNaN(serverParts[0]) && !isNaN(serverParts[1]);
 
-                // Compare major.minor (e.g. 10.8 vs 10.11)
-                var webMajorMinor = webParts[0] * 1000 + (webParts[1] || 0);
-                var serverMajorMinor = serverParts[0] * 1000 + (serverParts[1] || 0);
+                if (validSemver) {
+                    // Compare major.minor (e.g. 10.8 vs 10.11)
+                    var webMajorMinor = webParts[0] * 1000 + webParts[1];
+                    var serverMajorMinor = serverParts[0] * 1000 + serverParts[1];
+                    if (webMajorMinor >= serverMajorMinor) return;
+                }
 
-                if (webMajorMinor >= serverMajorMinor) return;
-
-                var minorDiff = (serverParts[1] || 0) - (webParts[1] || 0);
+                // Non-semver web version (e.g. "20220818.18-unstable") or web is behind server
                 var container = document.getElementById('webVersionMessage');
                 container.textContent = '';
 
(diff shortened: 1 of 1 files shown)


n00bcodr:
I think this is unnecessary complexity for a complete edge case? 

We cannot fix them all 😄

4eh5xitv6787h645ebv:
yep easy

PR #535, opened 2026-04-05
Update flag image source to use CDN for flags

Kisnov:
same as the past commit, update url to take all flags


n00bcodr:
Oversight, this should have been 

```js
flag.src = `https://cdnjs.cloudflare.com/ajax/libs/flag-icons/7.2.1/flags/4x3/${countryCode.toLowerCase()}.svg`;
```

Kisnov:
oh then I used the same link as the flags on the bottom left of posters, those are working tho, but should i change that to this also? I see you deleted "flagInfo".
On PR #464

PR #543, opened 2026-04-08
feat: add "Request More" button next to Series heading on Series pages

### LINKED ISSUE #502: ➕[Feature Request] Request new seasons on a Series' page.
**Is your feature request related to a problem? Please describe.**
Currently, if I need a new season, I should use the search bar to make a request. It works well since closing #409. Notably, one of my user needs accessibility and a virtual keyboard amd the search bar are a bit cumbersome.

**Describe the solution you'd like**
I think the UX could be better if I can directly goes to my Series page and request for new season. So a button "ask for new season" when the show misses at least, would be great, I think.
Actually, my user asked for it and I found relevant. So here I am sharing!

**Additional context**
Thanks for this amazing plugin! I have been using for months now my users love it!

4eh5xitv6787h645ebv:
## Description

Adds a "Request More" button next to the Seasons section heading (`h2.sectionTitle.sectionTitle-cards` inside `#listChildrenCollapsible`) on a TV Series detail page. Clicking it opens the existing Seerr season selection modal so users can request additional seasons without going through the search bar.

Motivated by issue 502: a user with accessibility needs relies on a virtual keyboard, and the search bar flow for requesting a new season of an already-in-library show is cumbersome for them. This button puts the same action one click away on the show's own page.

Request More button next to the Series heading

The button appears only when:
- Seerr integration is enabled and the user is connected
- The item is a Series with a TMDB ID but works with seerrs TVDB as well. Just needs the jellyfin to be TMDB.
- The show has at least one season that can be requested in Seerr (status 0/undefined/1/7)
- Admin has not disabled the new `JellyseerrShowRequestMoreOnSeries` setting (default: true)

## Changes

1. **New feature** (`item-details.js`, `PluginConfiguration.cs`, `JellyfinEnhancedController.cs`, `configPage.html`)
   - New `renderSeriesRequestMoreButton()` that runs alongside the existing `renderSimilarAndRecommended()` on every item-details navigation
   - Reuses the existing `checkForUnrequestedSeasons()` helper from `more-info-modal.js` (now exposed on `JE.jellyseerrMoreInfo`) so the detection logic stays in one place
   - Reuses the existing `JE.jellyseerrUI.showSeasonSelectionModal()` on click, so there is no new modal code
   - Button DOM is built entirely with `createElement` + `textContent` (no `innerHTML`)
   - Separate `AbortController` from the similar/recommended fetch so a slow recommendation lookup never cancels (or is cancelled by) this check on rapid navigation
   - Deduplication on both positive and negative paths to avoid repeating the expensive Seerr request lookup on every `viewshow` event
   - New admin setting `JellyseerrShowRequestMoreOnSeries` wired through the standard flow: PluginConfiguration default, public-config projection, configPage.html load/save

2. **Bug fix** (`more-info-modal.js`)
   - `checkForUnrequestedSeasons` was missing the `status === 7` (Deleted) case, even though the rest of `more-info-modal.js` treats status 7 as requestable (see lines 1214, 1701, 1752: `canRequestMain = !status || status === 1 || status === 7`)
   - Concrete repro before the fix: Futurama with all 10 main seasons in status 7 returned `hasUnrequested=false`, so neither the new Series-page button nor the existing in-modal Request More button would appear on it
   - This is a pre-existing inconsistency; the fix benefits the existing in-modal button too

3. **Reliability hardening** (`item-details.js`)
   - Replaced a `JE.helpers.onBodyMutation` subscription with a small `pollUntil` helper. The shared body observer is `{ childList: true, subtree: true }` with an explicit fast-path that drops dispatch on batches with no added/removed nodes, so it does not fire on the `classList.remove('hide')` and `span.textContent = 'Series'` mutations this waiter depends on. The code only appeared to work because Jellyfin happens to fire unrelated childList mutations around the same time
   - Added a `waitForChecker` helper that polls for `JE.jellyseerrMoreInfo.checkForUnrequestedSeasons` to become available before using it. `plugin.js` loads the Seerr modules in parallel via dynamically-inserted `<script>` tags (async by default), so on a cold page load `item-details.js` can execute before `more-info-modal.js` has attached its API. Without the wait, the button could be permanently missing until the user navigated away and back

## Implementation Notes

This PR was developed with AI assistance (Claude/GPT). All changes have been reviewed, tested, and understood.

The design intentionally reuses existing Seerr infrastructure rather than duplicating request logic:
- Detection: the existing `checkForUnrequestedSeasons` helper (now exposed)
- Click handler: the existing `showSeasonSelectionModal`
- Styling: the existing `.jellyseerr-request-button` / `.jellyseerr-button-request` classes, with a few inline overrides for inline placement inside an h2
- Heading layout: a small CSS class `.je-series-request-more-heading` injected at module init rather than mutating `heading.style` directly

## Testing

- [x] Built with `dotnet build` — 0 warnings, 0 errors
- [x] Deployed and tested on a Jellyfin 10.11.x dev instance
- [x] No console errors

**E2E matrix (Playwright on jellyfin-dev):**
- Big Bang Theory, Hazbin Hotel, Avatar: The Last Airbender → button appears (have unrequested seasons) and click opens the season selection modal
- Dexter, Halo, Fallout → button correctly skipped (all seasons already requested/available)
- Futurama → before the status 7 fix: button missing (all seasons in deleted state). After the fix: button appears, modal shows seasons ready for re-request
- Movie detail pages → button correctly absent
- Admin checkbox `JellyseerrShowRequestMoreOnSeries` loads, toggles, and persists across reload
- Cold-load race test: hard `page.goto()` directly on a Series detail URL with network throttled to 200 KB/s → button still appears, no `unavailable` warnings in the console
- Observer/poll lifecycle: registers and cleans up per navigation, no leaked timers or subscribers

## Config

New admin setting (in the Seerr → Item Details section of the plugin config page):

> Show "Request More" button next to Seasons heading on Series pages
>
> Adds a "Request More" button beside the Seasons section heading on Series detail pages when the show has unrequested seasons in Seerr. Lets users request additional seasons without going through the search bar.

Default: on. Disabling it hides the button without affecting the existing in-modal Request More button.

## Breaking Changes

None. The status 7 fix in `checkForUnrequestedSeasons` can cause the **existing** in-modal Request More button to newly appear on shows whose seasons are all in the deleted state — that is the intended behavior per the rest of `more-info-modal.js`'s requestability logic.

Closes #502


n00bcodr:
Thanks Jake! :)

PR #559, opened 2026-04-17
feat: multi-instance Sonarr/Radarr support

4eh5xitv6787h645ebv:
## Summary

Replaces the single Sonarr / Radarr URL + API key config with a list of `ArrInstance` records. Fixes issue 456.

- Multiple Sonarr/Radarr instances (TV vs Anime, SD vs 4K, etc.) can be configured from the plugin config page, each with its own name, URL, API key, optional per-instance URL mappings, and an Enabled toggle.
- All fan-out paths (arr links, Calendar page, Requests queue, scheduled Arr Tags Sync) query every enabled instance in parallel and merge results.
- Frontend renders a themed dropdown when an item exists in more than one instance, with per-instance status (complete / partial / missing), download count, size, and root-folder tooltip.
- Legacy single-instance fields are preserved as instance[0] on save for downgrade safety, and a missing `SonarrInstances` JSON automatically migrates from the legacy fields.

### Config UI

**Instance list (collapsed):** three Sonarr + three Radarr instances, each with an inline Enable checkbox + name + URL.
Collapsed instance list

**One card expanded:** body row is name input + Remove, followed by URL, API key + per-instance Test button, and collapsible URL Mappings.
Expanded card

**Disabled instance:** unchecking the summary checkbox marks the card (line-through name, "(disabled)" chip, dimmed background) without deleting credentials. Fan-out paths skip disabled instances.
Disabled instance

**New "Show status badge + color for single-instance links" toggle** (default off): keeps single-instance arr links as a plain icon/text when only one instance matches. Multi-instance dropdowns always show status.
Single-instance status toggle

### Item detail pages

**Single instance, default:** plain Sonarr icon — tooltip still surfaces full detail on hover (`Sonarr Main\n23/23 episodes\n32.8 GB\n/tv2`).
Single instance default

**Single instance, status enabled:** green border + `23/23` green badge → complete.
Single instance with status

**Multi-instance dropdown (Avatar: The Last Airbender — present in both Sonarr Main and Sonarr Anime (test)):** each row shows instance name, status color dot, episode badge (amber for partial, grey for missing), and size. Tooltip carries per-instance root folder.
Multi-instance dropdown

## AI assistance disclosure

This PR was developed with AI assistance (Claude / GPT). All changes have been reviewed, tested, and understood.

## Test plan

- [x] Built with `dotnet build Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj` → 0 warnings, 0 errors
[...]

PR #604, opened 2026-05-14
fix: admin config page UI bugs

### LINKED ISSUE #1: Prevent 'S' from reopening subtitle menu if it is open

4eh5xitv6787h645ebv:
## Summary

Bundles three small admin config page UI bugfixes:

1. **Radarr / Bazarr setup boxes can be configured independently of Sonarr.** The `*arr` tab was gating fieldsets[1] (Radarr) and fieldsets[2] (Bazarr) until a Sonarr URL+API key was configured, because `SECTION_DEPS` used `skipFirst: true` which only skipped fieldset[0]. Switched to a declarative `data-dep-setup` HTML attribute on setup fieldsets (Seerr Connection Setup, Sonarr, Radarr, Bazarr) so all four service-setup boxes are always editable. Dependent feature fieldsets (`*arr UI Links`, `Tags Sync`, Seerr feature fieldsets) still gate behind their `checkFn`s. **Note: I introduced this regression in my own PR #559 (multi-instance Sonarr/Radarr support).**

2. **Config page no longer overflows the viewport on portrait phones.** The mobile media query at `@media (max-width: 768px)` was setting `.jellyfin-tab-content.active { grid-template-columns: 1fr; }` without the `#JellyfinEnhancedPage` ID prefix, so its specificity (0,2,0) lost to the base rule's (1,1,0). On a Samsung S20 Ultra (412px) or iPhone (390px) the grid stayed pinned to a 480px minimum column, forcing every card 90-100px wider than the screen. Same specificity bug on `.jellyfin-tab-button` padding. Both rules now carry the prefix.

3. **Parent-dep hint text wraps below the label instead of beside it.** Parent-dep hints (e.g. `Enable "Enable Hidden Content" to configure`) are appended as direct `<div class="dep-hint-text">` children of `.checkboxContainer`. Since the container is `display: flex; align-items: center`, the hint became a flex sibling of the label, sharing the row. On narrow viewports both halves crushed into ~150px columns. Added `flex-wrap: wrap` to the container + `flex-basis: 100%` to direct-child hints so the hint falls onto its own line.

While in the `*arr` reactive code path for fix #1, also tightened a few things:

- Hook `debouncedUpdateDeps()` into the existing `wireRequestsBannerReactive` IIFE so the `*arr UI Links` / `Tags Sync` gate banners react in real-time as the admin types in a Sonarr/Radarr URL+key, instead of requiring save+reload.
- Tighten `hasAnyArrService()` to skip instances whose enabled-checkbox is unchecked. Matches the rest of the plugin's fan-out behavior (disabled instances are stored config but skipped by all controllers and scheduled tasks).
- Wrap `refresh()` body in try/catch so a thrown callback can't freeze the gate state under subsequent mutations.
- Surface `console.error` if `#JellyfinEnhancedForm` is missing instead of silently aborting the reactive wireup.

## Test plan

- [x] Built with `dotnet build` (0 warnings, 0 errors)
- [x] No new console errors
- [x] Playwright E2E (31 checks) for the *arr fix: Sonarr/Radarr/Bazarr setup boxes editable when no service is configured; `*arr UI Links` + `Tags Sync` gate banner visible initially; banner clears in real-time after typing URL+key; banner restores when enabled-checkbox unchecked; banner restores after instance removal; Seerr tab gating still works correctly
- [x] Playwright multi-viewport verification (40 checks across 8 viewports × 5 tabs) for the portrait/hint fix: 0px horizontal overflow at 375/390/393/412/768/820/844/1440px, desktop multi-column preserved (2 columns at 575px each on 1440px), parent-dep hint wraps below label on Pages tab

## AI assistance

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.css
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Configuration/configPage.css
@@ -1339,6 +1339,14 @@
     display: block;
     font-style: italic;
 }
+/* Parent-dep hints get appended as direct <div class="dep-hint-text"> children
+   of .checkboxContainer (which is display: flex; align-items: center). Without
+   these rules the hint becomes a flex sibling of the label, sharing the row
+   and crushing both pieces on narrow viewports (visible on phones: the label
+   "Use Plugin Pages for Hidden Content" and the hint "Enable X to configure"
+   end up squashed into half-width columns). Wrap the hint onto its own line. */
+#JellyfinEnhancedPage .checkboxContainer { flex-wrap: wrap; }
+#JellyfinEnhancedPage .checkboxContainer > .dep-hint-text { flex-basis: 100%; }
 .dep-required-icon { vertical-align: middle; margin-left: 6px; }
 
 /* ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


n00bcodr:
Included in v[11.10.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/releases/tag/11.10.0.0)

PR #630, opened 2026-05-24
Remove margin-right from letterboxd icon style

Macleykun:
Removed margin-right style from letterboxd icon.

Before:
image

After:
image

The bg being grey is from the theme i'm using :|

(extreme nit pick)
image
idk if you would also like the hover to match the logo color:
image

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/others/letterboxd-links.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/others/letterboxd-links.js
@@ -45,7 +45,6 @@
                     background-size: contain;
                     background-repeat: no-repeat;
                     vertical-align: middle;
-                    margin-right: 5px;
                 }
             `;
             document.head.appendChild(style);


n00bcodr:
Which theme is it you are using?

Macleykun:
> Which theme is it you are using?

GlassFin! But I also saw this with elegantFinn aswell! And through inspect element I though it was related to JE!

Macleykun:
> Which theme is it you are using?

Ah your right! It's indeed something related to the theme!

No extra padding:
image

With extra padding:
image

n00bcodr:
Yeah, it is dependent on the theme, I tailor these to vanilla Jellyfin and my own theme Jellyfish, But I do check if they work ok on other themes. 

But It is a bit difficult to get these working on every theme. :(

Macleykun:
> Yeah, it is dependent on the theme, I tailor these to vanilla Jellyfin and my own theme Jellyfish, But I do check if they work ok on other themes.
> 
> But It is a bit difficult to get these working on every theme. :(

Heyo! I looked into it and it's b/c of the repo that provides the icons! I made an issue there to suggest how it could be fixed. I'll close this PR tho as it's completly not relevant to JE! I do thank you for taking the time to reply!

PR #692, opened 2026-07-05
feat: serve third-party assets from a local CDN cache (+ off-thread tag-cache updates)

### LINKED ISSUE #660: ➕[Feature Request] Quality Tag reload or reset
It would be great if there were a button in the settings menu under “Quality Tags” that allowed you to reload or reset the quality tags.

I used to have this feature in another plugin, which I no longer use because I now have pretty much everything in Jellyfin Enhanced.

This feature is especially useful if you’re constantly swapping out movie or TV show files for example, from 720p to 1080p. Often, the 720p resolution still shows up for the movies or TV shows even after they’ve been replaced. That’s why I think a rescan of the files would be great.

Thanks for the awesome plugin :)

4eh5xitv6787h645ebv:
## Description

Fixes #660.

### 1. Local CDN subsystem — clients only ever hit a local plugin route

Every third-party **static** asset the client (and the admin config page) used to load
directly from an external CDN is now served from the plugin's own route, backed by an
on-disk cache that a scheduled task refreshes every 24 hours. Clients never contact an
external CDN host directly.

- **New route** `GET /JellyfinEnhanced/cdn/{source}/{**path}` (anonymous — these are
  public assets loaded via ``/CSS/`<link>`, which can't carry an auth token, and no
  user data is exposed), backed by **`Services/CdnAssetService.cs`**.
- **Fixed allow-list of upstream sources** (jsDelivr selfhst icons, Jellyfish theme
  sheets + logos, dashboard-icons, ZestyTheme logo, JellyPlugins favicon, Google Fonts
  Material Symbols font + css2 sheet, flagcdn + cdnjs flags, the ibb fallback poster, the
  Elsewhere region/provider lists, the Druidblack metadata-icon CSS, the plugin's own
  ratings CSS, doc screenshots, and newer-than-bundled locale JSON). A request path is
  strictly validated (no traversal/scheme/CRLF, printable chars only) and appended to a
  fixed base — the route can never be turned into an open proxy or SSRF vector.
- **Disk-backed cache** (not RAM) under
  `.../plugins/configurations/Jellyfin.Plugin.JellyfinEnhanced/cdn-cache/`, with a small
  in-memory hot layer, strong ETag/304, a 24h TTL, negative caching + an outbound-fetch
  concurrency cap on the anonymous route, and stale-copy fallback on upstream failure.
- The **Google Fonts css2** sheet is fetched and rewritten so its `@font-face` URLs
  resolve back to the local font route — the glyph files are proxied too.
- **New scheduled task** *Refresh CDN Assets* (`ScheduledTasks/RefreshCdnAssetsTask.cs`)
  runs on startup + every 24h, warming/refreshing the known static assets.
- Client side: a small `JE.cdn` helper in `plugin.js` (defined in the bootstrap so it's
  available before the parallel-loaded modules run) replaces every hardcoded CDN URL
  across `js/**` and `Configuration/configPage.html`.

Per-item TMDB poster/backdrop/provider-logo images (`image.tmdb.org`) and YouTube
trailer thumbnails (`img.youtube.com`) are intentionally left on their own hosts — they
are per-item content, not mutable CDN assets.

### 2. Process tag-cache updates off the library-scan thread

`ILibraryManager.ItemAdded/ItemUpdated/ItemRemoved` fire synchronously on the
library-scan thread, one item at a time. The tag-cache handler did heavy work inline
(`BuildEntryForItem` + media probe, plus a full Series/Season rebuild for every episode),
blocking large scans. Handlers now do only O(1) record-and-defer work; a debounced,
non-reentrant background worker coalesces changes by id and applies them off-thread,
persisting once (with `Dispose` drain + timer-resurrection guards). `WatchlistMonitor` and
`ContinueWatchingPlaybackEvents` move to the same record-and-defer pattern. Semantics are
unchanged — same items cached with the same data, only the timing/threading differs.

### 3. Reconcile tag-cache by Jellyfin item save time

The scheduled *Refresh Tag Cache* task now performs a delta reconciliation instead of
rebuilding every entry on every run.

- Persists `LastReconciledUtcTicks` in `tag-cache.json`.
- Queries Jellyfin with `MinDateLastSaved` so the database returns only taggable items
  saved since the previous successful reconciliation, with a small timestamp overlap.
- Rebuilds changed items only, plus parent Season/Series entries when an Episode changed
  because their quality tags can depend on first-episode stream data.
- Performs a cheap live-ID sweep to remove cache entries for deleted items without
  extracting stream data for the whole library.
- Falls back to a full build only when the cache is empty; non-empty legacy caches seed
  the reconciliation marker from the cache file timestamp and then run delta reconciliation.
- Serializes full builds and reconciliations with a `SemaphoreSlim` so manual and
  scheduled task runs cannot overlap.

This makes the scheduled task a safety net for the 720p -> 1080p replacement case: if the
library event path misses or races, Jellyfin's `DateLastSaved` still marks the changed
item for the next reconciliation.

## Benchmarks (measured on a dev instance)

**Tag-cache event cost** — a metadata edit fires the same `ItemUpdated` a library scan
raises, on the same handler and calling thread, so the edit's latency includes the
handler's inline cost. Measured on an episode of a 1230-episode series (One Piece), 25
timed edits, warm cache, only the plugin DLL changed. A control run reproduced the
numbers, ruling out drift:

| Build | Median per-edit latency (calling thread) |
|---|---|
| Before | ~72 ms |
| After  | ~42 ms |

The handler's inline contribution — ~30 ms per episode event — is removed from the calling
(scan) thread. Across that series' 1230 episodes a full scan raised this event ~1230
times, i.e. ~37 s of handler work that used to block the scan thread, now moved off it and
coalesced.

**CDN asset latency** — server fetch, warm cache (a real client also gains removal of the
internet round-trip and offline resilience):

| Asset | Before (external CDN) | After (local cache) |
|---|---|---|
| icon svg (~2 KB) | ~9 ms | ~3 ms |
| theme css (~1 KB) | ~9 ms | ~3 ms |
| flag png | ~20 ms | ~2.5 ms |
| Material Symbols woff2 (~330 KB) | ~677 ms | ~4 ms |

## Implementation Notes

This PR was developed with AI assistance (Claude/Codex). I have reviewed and tested all
changes and understand the implementation.

## Testing

- [x] Built for Jellyfin 10.11 (net9.0) and Jellyfin 12 (net10.0); deployed to both dev
      instances, plugin loads, *Refresh CDN Assets* task runs on startup (`Refreshed 32/32
      known assets`).
- [x] Rebuilt locally after the tag-cache reconciliation update:
      `dotnet build Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj -p:JellyfinTarget=jf10`
      and `dotnet build Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj -p:JellyfinTarget=jf12`.
- [x] `git diff --check`
- [x] Playwright E2E (admin **and** regular user): `JE.cdn` present, every source probe
      (icon/font/fontcss/flag/theme/locale) returns 200, local `/cdn/` requests observed
      with **zero** hits to any external CDN host; UI renders natively (flags, Material
      Symbols glyphs, sidebar plugin icons).
- [x] Admin config-page E2E: all 9 CDN images rewritten to the local route and actually
      loaded (naturalWidth > 0), font-face points at the local route, zero external hits.
- [x] Route hardening verified live: happy paths 200; bad source / path traversal -> 404;
      ETag -> 304; negative cache (2nd miss ~0 ms, no upstream); Google-Fonts CSS rewrite
      resolves through the local font route.
- [x] `node --check` on every edited JS file; translations validator green (567 keys,
      100%).
- [x] Multi-dimension review + adversarial verification; all confirmed findings fixed
      (anonymous-route DoS -> negative cache + fetch gate; hot-cache TTL; tag-cache timer
      resurrection after Dispose; concurrency races in the fixes) and false positives
      dismissed.


4eh5xitv6787h645ebv:
can split the pr if you want the tag cache and cdn 2 prs

n00bcodr:
> can split the pr if you want the tag cache and cdn 2 prs

Nahh, that's alright :)

PR #694, opened 2026-07-06
Fix MutationObserver leak in addUserPreferencesLink (one observer piled up per navigation)

### LINKED ISSUE #693: 🐛[BUG] MutationObservers from addUserPreferencesLink pile up on every navigation, UI gets slower over time
**Describe the bug**

The web UI gets progressively slower the longer a tab stays open (noticeable after ~1 hour of browsing). I profiled it and traced it to `addUserPreferencesLink` in `js/enhanced/ui.js`. It runs on every navigation (called from events.js), and unless the user-preferences menu is currently open, `addLinkToMenu()` returns false, so a new body-wide observer is created each time:

```js
if (addLinkToMenu()) return;
const observer = new MutationObserver(() => {
    if (addLinkToMenu()) observer.disconnect();  // only reachable while #myPreferencesMenuPage is visible
});
observer.observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['class'] });
```

During normal browsing the disconnect condition is never hit, so these stack up forever - I counted ~6 new observers per route change, 600+ after half an hour, each one notified on every DOM mutation on the page. Long-task time roughly doubled every 30 minutes; with just this code path stubbed out it stayed flat. Opening the preferences menu once (or F5) clears them, which fits the mechanism.

Quick way to see it - run this in the console, then navigate around:

```js
const Real = window.MutationObserver;
let created = 0;
window.MutationObserver = function (...a) {
  if ((new Error().stack || '').includes('addUserPreferencesLink')) created++;
  return new Real(...a);
};
window.MutationObserver.prototype = Real.prototype;
// `created` grows by ~6 per navigation and none of them ever disconnect
```

**Expected behavior**

One observer at most (disconnect the previous one before creating another), or route it through the shared body observer in helpers.js like the other components.

**Jellyfin Environment**

- Server: 10.11.11
- Plugin: 11.12.0.0 (the code path is unchanged on current main, ui.js line 568)
- Chrome (macOS), ElegantFin theme

**Additional context**

Happy to share the performance traces and heap snapshots I collected if useful.

007hacky007:
Fixes #693

## Problem

`addUserPreferencesLink()` runs on every `viewshow` and created a **new** `MutationObserver` on `document.body` with `{ childList: true, subtree: true }` each time, without ever disconnecting the previous one. Observers therefore accumulate one-per-navigation for the lifetime of the tab.

Each leaked observer:
- retains its callback closure (and detached DOM from the view it was created on), so heap grows monotonically, and
- keeps receiving callbacks for **every DOM mutation** under `document.body`, so per-mutation CPU cost grows linearly with navigation count.

On a long-lived session (TV dashboard, kiosk, HTPC) the web UI becomes noticeably sluggish after ~1 hour of normal browsing.

## Diagnosis

Found via Chrome DevTools heap snapshots + a `MutationObserver` wrapper tally on a real instance:
- Retainer chain: `MutationObserver` → callback closure → detached view DOM.
- Pre-fix: body-subtree observer count grows **+1 per navigation**, unbounded; heap grows monotonically and never returns to baseline after GC.

## Fix

Minimal, behavior-preserving: keep a module-level reference (`userPrefsLinkObserver`) and disconnect the previous observer before creating a new one, guaranteeing **at most one** live observer for this feature.

## Verification (patched build on a live server, 60 SPA navigations home ↔ item details)

| After navs | Observers created | Disconnected | Live | body-subtree observers | JS heap |
|---|---|---|---|---|---|
| baseline | 0 | 0 | 0 | 0 | 15.8 MB |
| 20 | 19 | 16 | 3 | **0** | 46.1 MB |
| 40 | 50 | 47 | 3 | **0** | 32.5 MB |
| 60 | 81 | 78 | 3 | **0** | 30.9 MB |

Body-subtree observer growth is gone (was +1/nav), live observer count is flat (the remaining 3 are jellyfin-web's own), and heap is GC'd back to baseline instead of growing. Full methodology and pre-fix numbers are in #693.

## Possible follow-up (out of scope here)

Even with the fix, one body-wide `subtree: true` observer stays active between navigations. A further optimization would be to disconnect it as soon as the link is injected, or observe a narrower container than `document.body`. Kept out of this PR to keep the diff minimal — happy to do it as a follow-up if you'd like.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/ui.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/ui.js
@@ -523,6 +523,8 @@
      * Injects the "Jellyfin Enhanced" link into the user preferences menu (mypreferencesmenu.html).
      * Adds it as the last item in the first vertical section (after Controls).
      */
+    let userPrefsLinkObserver = null;
+
     JE.addUserPreferencesLink = () => {
         const addLinkToMenu = () => {
             const menuContainer = document.querySelector('#myPreferencesMenuPage:not(.hide) .verticalSection');
@@ -564,14 +566,20 @@
         // Try to add immediately
         if (addLinkToMenu()) return;
 
-        // If not found, observe for when the menu is loaded and visible
-        const observer = new MutationObserver(() => {
+        // If not found, observe for when the menu is loaded and visible.
+        // Keep at most one pending observer: this function is called on every
+        // DOM change (see the dom-observer in events.js), and creating a new
+        // body-wide observer per call leaks observers until the menu is opened.
+        if (userPrefsLinkObserver) return;
+
+        userPrefsLinkObserver = new MutationObserver(() => {
             if (addLinkToMenu()) {
-                observer.disconnect();
+                userPrefsLinkObserver.disconnect();
+                userPrefsLinkObserver = null;
             }
         });
 
-        observer.observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['class'] });
+        userPrefsLinkObserver.observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['class'] });
     };
 
     /**


4eh5xitv6787h645ebv (approved):

n00bcodr:
Thank you! @007hacky007

PR #700, opened 2026-07-09
feat(spoiler-guard): per-user image identity tags for native clients

4eh5xitv6787h645ebv:
## Description

Adds **per-user image identity tags** to Spoiler Guard. Spoiler Guard already handles the opt-in blur and field-stripping behavior; this change improves request attribution for native-client image loads by stamping a stable per-user marker into image `tag` values returned in authenticated item DTOs. Clients echo that tag back on image requests, letting Spoiler Guard apply the correct user's blur state without depending on the request IP.

When Spoiler Guard and **Per-user image identity tags** are enabled:

- **Authenticated item DTOs** get image tags stamped with a stable per-user `-jeu{12hex}` marker across primary, backdrop, parent, series, logo, thumb, search-hint, recommendation, and chapter image tag fields.
- **Native image requests** can be attributed by the marker carried in `?tag=`, route `{tag}`, or `If-None-Match`, instead of relying on session-by-IP matching.
- **Reverse proxies / VPNs / shared NAT** no longer force unrelated users into the shared-IP fail-closed path when the client is using fresh marker-carrying image URLs.
- **Markerless requests** still fall back to the existing identity ladder, so old cached URLs and unsupported response shapes remain compatible.

The marker is not a credential. A forged marker can only make the caller opt into another user's Spoiler Guard view for media they could already request; it does not grant authentication or bypass Jellyfin authorization.

## Implementation notes

This PR was developed with AI assistance (Claude, Codex). The implementation was reviewed and tested in the local Jellyfin dev container.

- **Request identity ladder:** `Services/Identity/RequestIdentityService.cs` centralizes identity resolution: authenticated claims → image marker → single-user server → `je-spoiler-uid` cookie → shared-IP candidates → none. Consumers still choose their posture: field stripping requires one user, image protection can fail closed over all candidates.
- **Marker service:** `SpoilerIdentityService.cs` mints deterministic 12-hex SHA-256 user markers, resolves marker → current user, detects collisions, and invalidates its lookup map on user create/delete.
- **DTO stamping:** `SpoilerIdentityTagFilter.cs` stamps authenticated `BaseItemDto`, `SearchHintResult`, and recommendation response shapes. It also re-keys `ImageBlurHashes` so clients can still look blurhash placeholders up by the final stamped tag.
- **Filter ordering:** the MVC filter order keeps `SpoilerIdentityTagFilter` after `SpoilerFieldStripFilter` on the response unwind, so clients receive final cache-busted tags like `sb-{8hex}-{tag}-jeu{12hex}`.
- **Spoiler resolver split:** `SpoilerUserResolver.cs` now owns spoiler state loading / collection scope / corruption state, while request identity lives in `RequestIdentityService` so future features do not duplicate the IP/cookie/marker logic.
- **Cache invalidation:** `UserTopologyEvents.cs` invalidates single-user and marker caches when users are created or deleted; `UserConfigurationManager` invalidates cached spoiler state after successful `spoilerblur.json` writes.
- **Settings + docs:** adds the `SpoilerIdentityTags` admin setting, config-page load/save wiring, public config exposure, and docs for reverse-proxy-safe identity tags.
- **Translations:** no locale JSON changes were needed. The new visible setting text is in `Configuration/configPage.html`, which is static English like the surrounding plugin admin settings; client-side `js/locales/*` strings are untouched.

Notable robustness fixes:

- Marker extraction checks all carriers Jellyfin clients can send back: `?tag=`, route `{tag}`, and `If-None-Match` revalidation headers.
- The single-user shortcut avoids unnecessary session/IP scans on one-user servers but is invalidated immediately on user topology changes.
- Unknown, stale, deleted-user, or collision-excluded markers fall back to the previous cookie/IP ladder instead of misattributing requests.
- BlurHash dictionaries are re-keyed alongside image tags, including tags that already carry the `sb-{8hex}-` Spoiler Guard cache-bust prefix.
- Cross-request spoiler-state caching is invalidated after writes so image decisions do not serve stale protected/unprotected state for the cache TTL.

## Testing

Built locally and deployed the Jellyfin 10.11 artifact into the `jellyfin-dev` Docker container running **Jellyfin 10.11.10**:

- [x] `dotnet build Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj -p:JellyfinTarget=jf12`
- [x] `dotnet build Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj -p:JellyfinTarget=jf10`
- [x] `git diff --check`
- [x] Installed the `jf10` build into `jellyfin-dev` and confirmed Jellyfin loaded `Jellyfin.Plugin.JellyfinEnhanced, Version=11.12.0.0`
- [x] Confirmed JE startup completed after the plugin swap
- [x] Temporarily enabled `SpoilerBlurEnabled` for runtime testing and confirmed public config exposed `SpoilerBlurEnabled: true` and `SpoilerIdentityTags: true`
- [x] Confirmed authenticated item DTOs were stamped with per-user image markers
- [x] Confirmed two users receive different markers for the same DTO path: admin `f154c3824365`, Home `b3e50dc56670`
- [x] Confirmed marker stamping composes with cache-busted tags like `sb-131684bb-...-jeuf154c3824365`
- [x] Confirmed an anonymous image request using a stamped `?tag=...-jeu...` returned `200 image/jpeg`
- [x] Restored the dev container config back to `SpoilerBlurEnabled: false`; `SpoilerIdentityTags` remains true by default


4eh5xitv6787h645ebv:
fixed descriptions and removes the tags admin button

PR #722, opened 2026-07-28
feat(panel): searchable section navigation for the user settings panel

4eh5xitv6787h645ebv:
Reworks the user settings panel from the two-tab + stacked-`<details>` layout into an adaptive section view: a search-filtered nav rail on the left, one open section on the right, collapsing to a full-screen list/detail sheet on phones.

This is a port of the layout already shipping in [Jellyfin Canopy](https://github.com/4eh5xitv6787h645ebv/Jellyfin-Canopy), a JF12-only fork of this plugin — so the design has some mileage on it. It is a visual change to the most-used surface in the plugin, so the direction is yours to call.

## Why

The settings tab grew to 8 accordions in a panel that was `50vw` wide with `max-height: 90vh`. Finding a setting meant scrolling the accordion list, opening one, scrolling again, and there was no way to search. On a phone the fixed `50vw` inner width overflowed the panel and cards clipped horizontally.

## Before / After — desktop

| Before | After |
| --- | --- |
|  |  |

Shortcuts becomes the first section rather than a peer tab:

## Before / After — mobile (390×844)

| Before | After — section list | After — section open |
| --- | --- | --- |
|  |  |  |

Below 760px the panel becomes a full-screen sheet: the section list is the first screen and the detail pane covers it with a back button. The layer swap is deliberately instant — see [Canopy #513](https://github.com/4eh5xitv6787h645ebv/Jellyfin-Canopy/pull/513), where the equivalent `transition: transform 200ms ease` was measured overshooting its resting position and reading as a bounce. The transform that parks the closed pane off-screen stays; only the transition is gone.

## What changed

- `js/enhanced/ui.js` — panel markup and section-nav wiring:
  - each `<details>` becomes a `<section class="je-pane">` with an `.je-pane-title` heading; the shortcuts tab becomes the first pane
  - the nav rail is **built from the panes at runtime**, reusing each pane heading's markup (icon included), so nav and content cannot drift as sections are added or conditionally hidden
  - `#jePanelSearch` filters the nav by each pane's full text content, so searching "subtitle" also surfaces the Shortcuts section (it has a Subtitle Menu binding)
  - `lastOpenedTab` now stores a pane id and restores on desktop; phones intentionally open on the section list
  - Release Notes / Contribute / theme logo move out of the footer into a new **About** section, leaving the footer as just the close hint
- `js/locales/en.json`, `en-US.json`, `en-GB.json` — three new keys (`panel_search_placeholder`, `panel_back`, `panel_about_title`). Other locales fall back to English until Weblate picks them up.
- The user-preferences menu entry is renamed from **Advanced Settings (Jellyfin Enhanced)** to **Jellyfin Enhanced - User Settings** — the old wording read as a sub-page of Jellyfin's own settings rather than the plugin's user panel. Left as a plain literal, matching how it was before.

### Two behaviour fixes the new layout forced

- **Drag is now scoped to the header.** `handleMouseDown` skipped drags originating in `.preset-box, button, a, details, input`. With `<details>` gone, that exclusion stopped matching the settings markup, so a blanket panel-drag stole pointer gestures from the subtitle colour sliders. It now only starts from `.panel-header`.
- **`?` no longer closes the panel while typing.** `?` is a close shortcut, which made the new search box unusable the moment you typed a question mark. `closeHelp` now ignores it when the event target is a text field or contenteditable; `Escape` and `?` outside a field are unchanged.

Phone mode also sets `inert` on whichever of the two stacked layers is hidden, so the offscreen layer stays out of the tab order and the a11y tree, and focus moves to the back button on drill-in and back to the active section on drill-out.

## Testing

Built for both targets — Jellyfin 10.11 (`-p:JellyfinTarget=jf10`) and Jellyfin 12 (`-p:JellyfinTarget=jf12`), 0 warnings each.

The checks below were driven with Playwright on a clean 10.11.11 container, with no console errors:

- desktop restores the last-opened section; phone opens on the list
- search filters the rail; clearing it restores all 8 sections
- typing `?` in the search box keeps the panel open; `?` outside a field still closes it
- toggling Auto-resume round-trips to the server (`false → true` persisted)
- phone drill-in/back, and `inert` ownership matching the visible layer
- pane motion, frame-sampled at 390×844: `transitionDuration` is `0s` and the pane jumps straight between its two end states, with the closed pane still parked fully off-screen

```
open:  397.8, 397.8, 397.8, 397.8, 0, 0, 0, 0, ...
back:  0, 0, 0, 0, 397.8, 397.8, 397.8, 397.8, ...
```

The jf12 artifact was then deployed to a live Jellyfin 12.0.0 server and the redesigned panel confirmed working there too, on the modern (React/MUI) layout — so this lands on both supported runtimes.

`node scripts/validate-translations.js validate en-US` / `en-GB` both pass at 100%.


n00bcodr:
Available in version [12.1.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/releases/tag/12.1.0.0)

PR #723, opened 2026-07-31
refactor: modularize the client by feature

4eh5xitv6787h645ebv:
## Description

Opened as a **draft** for discussion — this is a large structural change and I would rather agree the shape before it is treated as ready.

Splits the five client-side monoliths into per-feature modules and adds a shared `js/core/` layer.

**JavaScript, its documentation, and five stale code comments.** Every other non-JS path — `configPage.html`, locales, workflows, `manifest.json` — is byte-identical to `main`. Outside `js/` this PR touches exactly five files: `docs/advanced/project-structure.md` and the *Project Structure* section of `CONTRIBUTING.md` (both described the pre-modular tree), plus five comment lines across `JellyfinEnhancedController.cs`, `RequestIdentityService.cs` and `configPage.css` that pointed at monoliths the split retired. **No behaviour changes outside `js/`** — every `.cs` and `.css` line in the diff is a comment. A scan of all deleted/renamed module names across every non-`js` file type confirms no stale reference remains.

| | main | this branch |
|---|---:|---:|
| JS files | 63 | 146 |
| Total JS lines | 53,577 | 53,622 (**+0.08%**) |
| Average file size | 850 | 367 |
| Largest file | 3,556 | 1,227 |
| Registered modules | 59 | 142 |

The five monoliths: `more-info-modal.js` 3,556 → 8 files · `calendar-page.js` 2,997 → 6 · `jellyseerr/ui.js` 2,897 → 10 · `enhanced/ui.js` 2,648 → 9 · `requests-page.js` 2,628 → 7.

The split added **+45 lines across the whole client** — it introduced essentially no duplication, because it was funded by real deduplication.

### `js/core/` is the substance

2,141 new lines (`api-client.js` 636, `tag-renderer-base.js` 468, `navigation.js` 318, `dom-observer.js` 270, `ui-kit.js` 229, `lifecycle.js` 220) paid for by `enhanced/helpers.js` **989 → 438** and `jellyseerr/request-manager.js` **438 → 15** (now an alias shim).

- **Navigation** — ~36 ad-hoc `hashchange`/`viewshow` listeners that double-fired on hash navigation and missed `pushState` navigation are now one deduped dispatcher.
- **Tag rendering** — `isCardAlreadyTagged`/`markCardTagged` were duplicated per tag module; now `TAGGED_ATTR` + `ctx.isTagged()` in `core/tag-renderer-base.js`, used by 4 of 6 renderers.
- **Discovery** — `discovery-base.js` owns pagination (3 strategies), abort, dedupe, filtering, lifecycle. All 5 discovery modules use it, none bypass it; a new row is a ~12-line spec.
- **UI primitives** — one `escapeHtml` (previously defined 3+ times), one deduped CSS injector, one toast, one scroll-friendly tap handler.

### Why this beats the monolith

- **Cross-feature fixes land at the source.** The touch-scroll fix put `addTouchTapListener` inside `jellyseerr/ui.js` on `main`. Here it is in `core/ui-kit.js`, byte-identical, shared by the results icon, cards and request buttons.
- **Reviewable diffs.** A CSS change no longer drags 3,000 lines of context; `*-styles.js` isolates presentation churn.
- **It already caught a live bug in `main`.** Commit `454ca95` deleted `const isPlainLeftClick` but left its use at `jellyseerr/ui.js:1402`, so plain left-clicking an external Seerr link on a search card throws a `ReferenceError`. The definition survives here (`jellyseerr/ui/ui-cards.js:293`). That is a separate one-line fix `main` should take regardless of this PR.

## Implementation Notes

**This PR was developed with AI assistance (Claude).** All changes have been reviewed and tested, and the implementation is understood — see the verification matrix below, which was run against live servers rather than asserted.

**This is not pure code motion and cannot be verified by move-detection alone:**

- **32 top-level function names on `main` do not exist by name here.** Every one checked is a rename or absorption into a shared abstraction — `getAuthHeaders` → `core/api-client` `authHeaders()`; the four `get*IdFromUrl` helpers → `idFromDetailUrl`/`idFromListParam`; `render*Discovery`/`fetch*Discover` → the `createDiscovery` spec; `buildSpoilerToggleButton` → `spoilerguard/seerr-toggle.js` `buildSeerrToggle`; `isCardAlreadyTagged` → `core/tag-renderer-base.js`. No behaviour was intentionally dropped, but the burden of proof sits with this PR.
- **`core/navigation.js` patches `history.pushState`.** `main` already does something similar more crudely, but this is the highest-risk file here and warrants separate scrutiny; it could reasonably land as its own prior PR.

### Security

Per `.github/SECURITY_GUIDELINES.md` §4 (XSS Prevention):

- `innerHTML` occurrences: **196 → 197** (net +1, pure code motion; +139/−138 across the diff).
- `escapeHtml` call sites: **184 → 206** — escaping increased, not decreased.
- No new `eval`, no new user-input sink, no auth/permission logic changed (all C# is untouched).
- **No new executable `innerHTML` sink**: counting only non-comment lines, the diff adds 127 and removes 127. The +1 in the raw total is a comment/JSDoc mention; the one relocated sink is the toast, moved from `enhanced/ui.js` to `core/ui-kit.js` with its parameter renamed `key`→`html`.

### Breaking changes

None for users. For contributors: adding a JS file now requires appending to `allComponentScripts` in `plugin.js` **after** every module whose exports it reads at load time.

## Testing

Fresh throwaway containers, plugin built per target (`jf10` → net9.0, `jf12` → net10.0):

- [x] No console errors
- [x] Compatible with Jellyfin 12 — tested on **12.0.0**
- [x] Mobile compatibility — 390×844, phone sheet + `inert` layer handover verified
- [x] Chrome/Chromium and Firefox
- [ ] Edge — not tested directly (Chromium engine, same as Chrome)
- [ ] Safari/WebKit — not tested (browser unavailable in this environment)

| | 10.11.11 | 12.0.0 |
|---|---|---|
| Plugin status | Active | Active |
| Modules returning 200 | 142 / 142 | 142 / 142 |
| `failedModules` | none | none |
| Failed `/JellyfinEnhanced/` requests | 0 | 0 |
| Page errors | 0 | 0 |
| Console errors | 0 | 3 — all Jellyfin's own `ws://…/socket` 403 |
| Settings panel (desktop) | 8 panes → 8 nav items, search filters, Esc closes | identical |
| Settings panel (mobile) | 390px sheet, list-first, `navInert` on open, Back restores | identical |
| Chromium / Firefox | both clean | both clean |

### Benchmark vs `main`

Cold cache each iteration, median of 9, measured from navigation to the `All component scripts loaded` console event — the only readiness signal both builds emit.

| | requests | payload | 10.11.11 | 12.0.0 |
|---|---:|---:|---:|---:|
| `main` (monolith) | 59 | 2,321,562 B | **1058 ms** | **1055 ms** |
| this branch | 142 | 2,321,702 B | **1057 ms** | **1250 ms** |
| delta | 2.41× | +0.006% | −1 ms (0%) | **+195 ms (+18%)** |

**This is not a performance PR, and the split is not faster.** It is byte-neutral (140 bytes across the entire client), but 59 → 142 requests is free on 10.11 and costs ~195 ms on 12.0.0, where the distributions barely overlap (branch 1210–1437 ms vs main 786–1260 ms). `script.async = false` keeps downloads parallel, so this is per-request overhead, not serialisation. The lever, if it matters, is bundling for delivery while keeping split sources — the plugin controls its own script route. Deliberately not in this PR.

### Known costs, stated plainly

- **Per-change fan-out is ~3×.** The searchable-settings-panel change touched 1 file on `main`; here it touches 5. Requests history: 1 → 5.
- **Correctness moved from closure-guaranteed to order-dependent.** ~290 load-time captures mean a misordered entry in `allComponentScripts` binds `undefined`, and the codebase's defensive `?.`/`typeof` guards hide it — the feature silently never appears instead of throwing. Nothing validates the array.
- **That failure mode already bit.** 37 of 142 modules (26%) were silently 404ing: MSBuild rewrites a hyphen in a *directory* segment of an embedded-resource name to an underscore while leaving *file* hyphens alone, so `js/enhanced/settings-panel/ui-panel.js` embeds as `…settings_panel.ui-panel.js` and the controller's `path.Replace('/', '.')` lookup missed it. Fixed by renaming the five hyphenated directories to the concatenated form already used by `spoilerguard` — no C# change. Called out because it is the clearest evidence of the trade-off above.
- **Upstream syncs become re-derivations.** Every upstream edit to a retired monolith is a modify/delete conflict resolved by hand, and upstream churn concentrates on exactly the most-split files.

### Granularity

142 registry entries is a lot. Modules that arguably do not justify a file: `jellyseerr/request-manager.js` (15 lines, alias shim), `spoilerguard/ids.js` (30), `spoilerguard/identity.js` (40), `jellyseerr/ui/ui-icons.js` (43). Spoiler Guard is 12 files / 958 lines replacing one 977-line file. These can be coarsened before merge.


n00bcodr:
Available in version [12.1.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/releases/tag/12.1.0.0)

PR #738, opened 2026-08-05
fix(player): harden OSD track cycling (waits for the menu, stops matching English UI text)

### LINKED ISSUE #705: 🐛[BUG] "Cycle Subtitle Tracks" keyboard shortcut doesn't work
**Describe the bug**
Pressing the "Cycle Subtitle Tracks" keyboard shortcut (`C` by default) during playback shows a toast notification saying "No Subtitles Found", despite Bazarr being properly configured and subtitles being suggested when clicking the "CC" button.

**Expected behavior**
Pressing the "Cycle Subtitle Tracks" keyboard shortcut cycles through available subtitles.

**Screenshots**
Image
Image

 - Server Version: 10.11.11
 - Plugin Version: 11.12.0.0
 - Jellyfin Installation Method: Linux
 - Device: Jellyfin Web on Firefox
 - External Access Method: None
  - Theme being used: [e.g. Jellyfish, Zesty, etc.,]
 - [x] [FileTransformation](https://github.com/IAmParadox27/jellyfin-plugin-file-transformation) Installed

**Logs**
None relevant.

**Additional context**
None.

4eh5xitv6787h645ebv:
## Description

Hardens the OSD subtitle/audio track cycling (`Cycle Subtitle Tracks` / `Cycle Audio Tracks`) — the code path behind #705.

**The #705 symptom is reproduced deterministically on Firefox with the released plugin, and this branch fixes it** (see the Firefox section below). On Chromium the symptom does not occur; there the PR fixes a set of related confirmed defects in the same function — localization-dependent sheet stacking, wrong-track selection, and a timing race. I've left the maintainer to decide whether to link this as closing #705, since the reporter's browser is unknown. See *Honest scope* below.

## The defects

All verified against the `jellyfin-web` 10.11 and 12 sources. Both lines ship the same legacy video OSD (`controllers/playback/video/`), and `actionSheet` / `dialogHelper` behave identically between them, so every item applies to both.

**1 — The action sheet was read after a blind 200 ms delay.** Nothing verified it had opened. The sheet is built inside a dynamic `import()` of `components/actionSheet`. Now waits for the sheet via `MutationObserver`.

**2 — Queries were document-wide, so they could read a dismissed or stacked sheet.** `dialogHelper.close()` only *hides* a sheet after a ~100 ms exit animation and removes it from the DOM later still, so `document.querySelectorAll('.actionSheetContent .listItem')` can match a sheet the user already dismissed — or two sheets at once. Every query is now scoped to one resolved `.dialogContainer`.

**3 — `document.body.click()` cannot close a Jellyfin action sheet.** Three independent reasons: the close listener is bound to `.dialogContainer`, which is a *child* of `body`, so a body-dispatched event never reaches it; the handler requires `e.target` to **be** that container; and it also requires a preceding `mousedown`, while `HTMLElement.click()` dispatches only a `click`. All five call sites were no-ops.

**4 — The "already open" fast path compared the sheet title to the literal English `'Subtitles'` / `'Audio'`.** Those titles are localized, so on a non-English UI the branch never matched: the code re-clicked the button and **stacked a second sheet**, after which the document-wide query read rows from both. Reproduced deterministically on a German UI (1 sheet became 2; the row query saw 66 rows instead of 33; the wrong track was selected). The code now compares against the OSD button's own `title` attribute, which resolves the same translation key — locale-proof.

**5 — Track rows were filtered by excluding the literal `'Secondary Subtitles'`, which is also localized.** On a localized UI that row was not filtered out, so the cycle could land on it and click it — opening the *secondary subtitle* submenu instead of changing the track. Rows are now identified by their numeric `data-id` (the stream index; `-1` is *Off*), which is language-independent.

**6 — Cycling could interfere with the plugin's own long-press 2× feature.** The long-press handlers are capture-phase document listeners; the synthetic close events the new code dispatches on a `.dialogContainer` would have armed the press timer and had the click cancelled before `dialogHelper`'s own listener saw it — leaving the sheet open and playback stuck at 2× when the timer fired. Long-press now ignores untrusted (synthetic) events, which also protects it from any other programmatic clicks.

**7 — Rapid repeated presses could re-select the same track.** Jellyfin applies a selection only when the sheet's close animation finishes (`actionSheet` resolves on dialog close, not on the row click), so for a beat after a cycle the check mark still sits on the previous track. The cycle now remembers the row it just clicked (per menu kind, scoped to the current item, short window) and continues from there; sheets that are mid-close are also excluded from reuse via `dialogHelper`'s own `closing` event. Key auto-repeat is ignored for the two cycle shortcuts — now that every press is a real stream switch, a held key would otherwise fire a ~30 ms burst of track changes (transcode restarts, for burned-in subtitles or audio).

## The fix

- Wait for the sheet with a `MutationObserver` instead of a fixed delay, and resolve only on a container that did not exist before the click **and** whose title matches the menu requested.
- Scope every subsequent query to that container.
- Match the "already open" case against the **OSD button's own `title` attribute**, which resolves the same translation key as the sheet title.
- Identify track rows by their numeric `data-id`; a menu whose only row is *Off* reports "no subtitles" instead of toggling Off→Off with a track-changed toast.
- Close sheets by dispatching `pointerdown` → `mousedown` → `click` on the dialog container (the sequence `dialogHelper` and the Firefox/Edge OSD click guard both require — `JE.skipIntroOutro` already uses the same pattern).
- Guard against re-entrancy (a second press before the sheet exists no longer stacks an orphan sheet), navigation away mid-wait, and empty menus.

`core/dom-observer.js`: `waitForElement` gains `{ timeout, predicate, root, quiet }`, staying backward compatible with a numeric timeout. The `predicate` is what makes it usable here — it previously resolved immediately whenever the selector matched, and a dismissed sheet always matches. The `Date.now()` observer-id suffix also became a counter, since two waits starting in the same millisecond collided.

## The Firefox cause of #705 — now reproduced and confirmed fixed

On Firefox/Edge only, the video OSD installs a document capture-phase click guard (`video/index.js` `onClickCapture`) that cancels any click whose target does not contain its `clickedElement`. That variable is refreshed by `keydown` and by `pointerdown` — **not** by `mousedown`. A programmatic `.click()` issued from a keyboard shortcut therefore has `clickedElement` pointing at the keydown target, and gets swallowed.

That produces exactly #705's symptom, and it is **no longer a theory**. On a clean Jellyfin 10.11.11 with a movie carrying 3 subtitle + 2 audio tracks, driven by Playwright Firefox with real key input:

- **Current released plugin:** every press of the cycle shortcut toasts "❌ No Subtitles Found"; no menu ever opens. The #705 symptom, deterministic, on a video with subtitles.
- **This branch:** the same presses cycle English → German → French → Off → English (and audio English ↔ German), one advance per press, correct wrap through *Off*.
- A rapid double-press 250 ms apart — inside the window where Jellyfin has not yet applied the first selection — advances **two distinct tracks** (German, then French) and leaves zero orphan sheets.

The fix is the same `pointerdown`-before-click treatment `JE.skipIntroOutro` already uses for this exact guard, applied to the OSD button and the menu row. If #705's reporter was on Firefox or Edge, this is very likely their bug.

## Honest scope

- **Reproduced and fixed:** the localization defects (4, 5) and the sheet stacking they cause.
- **Confirmed from source and fixed:** 1, 2, 3, 6, 7.
- **Not reproduced:** the 200 ms race itself. Rows render in ~47 ms even for an item with 32 subtitle streams, and no network request fires at that moment — `elements/emby-select` *statically* imports `actionSheet`, so that chunk is usually already loaded. That is very likely why #705 has been hard to reproduce. The delay is still unsound and is now gone, but I make no claim it was the cause.
- **Reproduced on Firefox and fixed:** the click-guard swallow above — the released plugin shows the exact #705 symptom on Firefox; this branch does not.

## Testing

- [x] No console errors
- [x] Compatible with Jellyfin 12 — tested on **12.0.0**
- [x] Mobile compatibility — 390×844
- [x] Chrome/Chromium
- [x] Firefox — #705 symptom reproduced with the released plugin and confirmed fixed with this branch (subtitle + audio cycling, rapid double-press, no orphan sheets)
- [ ] Edge / Safari — not tested

The first two commits were verified on live Jellyfin 10.11.11 and 12.0.0 servers with a real 962-item subtitled library: the German repro reuses the open sheet instead of stacking (correct track selected), English scenarios behave identically before and after, and the settings-panel/module matrix stays clean. The third commit (defects 6 and 7 plus hardening) is source-verified against `jellyfin-web` 10.11 and master, passed two independent adversarial review rounds, and its cycling behavior (including the rapid double-press path) was runtime-verified in the Firefox matrix above.

## Implementation Notes

**This PR was developed with AI assistance (Claude).** The changes have been reviewed and tested, and the reasoning above is grounded in the `jellyfin-web` sources rather than asserted — including the parts I could not confirm, which are labelled as such.

### Breaking changes

None.

Generated with Claude Code & Canopy.


n00bcodr:

PR #739, opened 2026-08-05
fix(tags): preserve explicit language regions in flag resolution (pt-BR, es-419, zh-Hant…)

### LINKED ISSUE #544: ➕[Feature Request] Language Flag Change
I know it will sound a bit like unimportant, but it's just to make it easier for people to know like, if the content is dubbed in a language, which it is. Portugal and Brazil's accent and dubbing scene are vastly different, for example, so having a Portuguese flag to say Portuguese and not knowing which is kinda bad. I would like to be able to have at least some control in telling: Oh, this Portuguese is from Portugal, this is from Brazil in the language context, same for different Spanish from Latin America, or different English, and so on.

I know it sounds like a mini thing, but get a video of both Portugal and Brazilian speaking and see how afar they sound.

4eh5xitv6787h645ebv:
## Description

Language flags now respect explicit region variants. An audio track tagged `pt-BR` gets the Brazilian flag while `pt` / `pt-PT` keep the Portuguese one; `es-419` / `es-MX` / `es-LA` (Latin-American Spanish) get the Mexican flag while `es` stays Spain; `zh-Hant` maps to the Taiwanese flag; the de-facto Brazilian Portuguese codes `pob`/`pb` are understood. A bare base language is never guessed into a regional variant.

Closes #544

## Screenshots

Test movie with three audio tracks: `pt-BR`, `es-419`, `por` (Jellyfin 10.11.11, default theme).

| | Before | After |
|---|---|---|
| **Poster overlay** | before: single Portuguese flag | after: Brazilian, Mexican and Portuguese flags |

**Details page, before** — the regioned tracks get no flag at all, only bare `por` does:

before: details row with flags missing

**Details page, after** — every track gets its regional flag:

after: details row with Brazilian, Mexican and Portuguese flags

## What was wrong

- The Language Tags overlay stripped the region subtag during normalization (`code.split('-')[0]`) in two places, so every regional variant collapsed to the base language's flag.
- The details-page audio row kept the full code but then looked it up in a map that has no regioned entries — and its display name ("Brazilian Portuguese") matched nothing either — so regioned tracks got **no flag at all** there.
- The two features each carried their own copy of the language→country map, already drifting.

## The fix

- New shared module `js/core/media-language.js`: one map + one region-aware resolver used by both features. Resolution order: explicit valid region (alias-normalized: `en-UK`→gb, `es-LA`→mx; validated against the ISO 3166-1 set so a private-use region like `pt-XA` falls back to the base flag instead of a broken image) → Chinese script subtag (`zh-Hant`→tw) → base-language default → display-name lookup.
- Guards that came out of review: the curated Catalan/Galician/Basque subdivision flags win over the generic Spain flag for `ca-ES`/`gl-ES`/`eu-ES`; BCP-47 extension sections are never misread as regions (`en-u-ca-gregory` stays gb); script mapping is Chinese-only (`ja-Hans` stays jp); numeric regions are per-language conventions, not global (`en-419` stays gb); a region on an unmapped base (`und-419`) stays flagless.
- Overlay dedupe is now per flag, so a movie carrying both `pt-PT` and `pt-BR` tracks shows both flags (previously impossible).
- The overlay's localStorage cache key is bumped (`-v2`) so entries cached with stripped codes can't pin the wrong flag for the 30-day TTL; the renderer base's legacy-key cleanup now accepts a list of stems so both prior key generations keep getting removed.
- The pt-BR mapping concern raised in the issue thread is covered: `pob`, `pb`, `pt-BR`, `pt_BR` and the "Brazilian" display name all resolve to the Brazilian flag.

Server-side extraction already preserved regions — the loss was purely client-side, so this is a JS-only change.

## Testing

- [x] No console errors
- [x] Compatible with Jellyfin 10.11.x — verified live on **10.11.11**
- [x] Doesn't break existing functionality — a `por`+`spa` control movie renders identically before/after
- [x] Chrome/Chromium and Firefox (Playwright, real UI injection)
- [x] Jellyfin 12 build compiles; the change is identical shared JS (not separately runtime-tested on 12)
- [x] 50-case resolver matrix (regions, aliases, scripts, extensions, junk, legacy cache shapes) passes
- [x] `mkdocs build --strict` and translation validation pass (no new strings)

Live verification: a movie with `pt-BR` + `es-419` + `por` audio renders Brazilian, Mexican and Portuguese flags on its poster overlay and its details-page audio row, with correct "Brazilian Portuguese" / "Latin American Spanish" tooltips from `Intl.DisplayNames`.

## Implementation Notes

**This PR was developed with AI assistance (Claude).** The changes went through two adversarial review rounds; every confirmed finding (including a regression the first draft introduced for `ca-ES`) was fixed and re-verified.

### Breaking changes

None. Media tagged with bare base languages renders exactly as before; the language-tags localStorage cache rebuilds once under the new key.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/core/media-language.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/core/media-language.js
@@ -1,0 +1,218 @@
+// /js/core/media-language.js
+// Shared audio-language → flag resolution for the Language Tags overlay
+// (js/tags/languagetags.js) and the details-page audio-language row
+// (js/enhanced/itemdetails/features-details-media-info.js). Single source of
+// truth so the two features can never disagree about which flag a language
+// gets.
+//
+// The resolver understands region subtags: a stream tagged `pt-BR` gets the
+// Brazilian flag while `pt` / `pt-PT` keep the Portuguese one, `es-419` and
+// `es-MX` get the Mexican flag while `es` stays Spain, `zh-Hant` maps to the
+// Taiwanese flag, and so on. A region flag is shown only when the media
+// explicitly carries one — a bare `pt` or `es` is never "guessed" into a
+// regional variant.
+(function(JE) {
+    'use strict';
+
+    JE.core = JE.core || {};
+
+    /**
+     * Language name/ISO-code → flag code, for languages WITHOUT an explicit
+     * region subtag. Values are flag-icons codes (ISO 3166-1 alpha-2, plus
+     * subdivision codes like `es-ct` which the flag-icons set also ships).
+     * Keys cover English display names (as produced by Intl.DisplayNames) and
+     * ISO 639-1/639-2 codes as they appear in real-world media containers.
+     */
+    const baseLanguageFlags = {
+        English: 'gb', eng: 'gb', en: 'gb', Japanese: 'jp', jpn: 'jp', ja: 'jp', Spanish: 'es', spa: 'es', es: 'es',
+        French: 'fr', fre: 'fr', fra: 'fr', fr: 'fr', German: 'de', ger: 'de', deu: 'de', de: 'de',
+        Italian: 'it', ita: 'it', it: 'it', Korean: 'kr', kor: 'kr', ko: 'kr', Chinese: 'cn', chi: 'cn', zho: 'cn', zh: 'cn',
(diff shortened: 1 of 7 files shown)


eerison:
Hey @4eh5xitv6787h645ebv 

Thanks for this feature, but I am using the latest version, and it keep showing Portugal flag only. Should I do something else?

n00bcodr:
@eerison a plugin with that change has not been released yet, will release probably during the weekend.

n00bcodr:

eerison:
Should I clean some cache? I installed 12.2.0 and restarted via ui.

But I can't see any difference.

PR #740, opened 2026-08-05
fix: re-scope all per-user state when switching users without a page reload

4eh5xitv6787h645ebv:
## Fix stale per-user data after switching users in the same browser

Fixes the cache/user-switch problem behind the *"My Bookmarks page is viewable to all users"* discussion, the *"Radarr/Sonarr/Bazarr buttons can appear for non admin users"* bug report, and completes the cache fix that came out of *"Jellyfin enhanced settings common to all users"* (the earlier cache-busting fix covered the HTTP layer but not the in-memory state). Issue links to follow when this is ready for review.

### Problem

Jellyfin's web client is an SPA — logging out and back in as a different user never reloads `index.html`. Jellyfin Enhanced bootstraps once per page load and snapshots the signed-in user into module-level state (`JE.userConfig`, `JE.currentUser`, `JE.currentSettings`, plus ~25 per-feature caches). After a user switch, all of it still belongs to the previous user until a hard reload (Shift+F5). The most visible symptom was the Bookmarks page via Custom Tabs showing the previous user's bookmarks, but the same staleness affected Seerr identity/permissions, Spoiler Guard state, hidden content, watch progress, tag caches, admin-gated UI (arr links, active streams), the pause screen's captured token, and several localStorage keys.

### Fix

**New `js/core/session.js`** — an identity-epoch tracker:
- Wraps `ApiClient.setAuthenticationInfo` (prototype-chain walk, idempotent, only reacts to the active client) so the transition runs the moment credentials change; the shared `je:navigate` pipeline and a slow reconcile interval are belt-and-braces fallbacks.
- Every transition bumps a monotonic **epoch** and synchronously runs registered reset handlers (`JE.session.onUserChange(name, fn)`), then emits `je:user-changed`.
- Async loaders capture the epoch before awaiting and drop results if the user changed mid-flight (`JE.session.isCurrent(epoch)`), closing the "user A's slow response lands in user B's session" race.

**`plugin.js` re-bootstrap** — on a switch, the per-user globals are wiped, then the incoming user's five config files, user object, admin-only private config, settings, shortcuts and translations are re-fetched (epoch-guarded), and `je:user-data-loaded` re-renders per-user views (bookmarks, hidden content, tags, arr links, active streams) without a reload. A switch that happens mid-boot (before session.js loads) is also detected and repaired.

**Per-feature reset handlers** across every module that holds per-user state, plus epoch guards on every fetch continuation that writes a cache — including debounced/retry save timers (hidden content) that could otherwise erase the incoming user's server-side file.

**Storage hygiene**
- The frozen tag-cache localStorage key NAMES are preserved; a sibling `…:identity-owner` sentinel records which server:user wrote the payload and drops it on mismatch.
- `je_hide_confirm_suppressed_until` and `je.calendar.showUnmonitored` are now per-user (calendar's legacy value is adopted once).
- The `je-spoiler-uid` cookie is rewritten/cleared on every switch.
- Server: `Cache-Control: no-store` on all identity-varying GET endpoints (`user-settings/*`, `tag-cache`, `private-config`, `jellyseerr/user-status`, `spoiler-blur/series`, `arr/requests`).

### Compatibility

Works on Jellyfin 10.11.x and 12 — the mechanism relies only on `ApiClient.setAuthenticationInfo`, the existing `je:navigate` pipeline and plain property reads, no web-client internals.

### Testing

- Automated single-document account-switch E2E (Playwright) on fresh Jellyfin **10.11.11** and **12 (unstable)** containers: login A → UI logout → login B with reload banned (verified via a window marker + `performance.timeOrigin`). Asserts: A's bookmarks wiped at logout, B's bookmarks (and only B's) loaded after switch, `JE.currentUser`/session epoch/spoiler cookie all track the switch. Both pass.
- `node --check` on all changed JS; jf10 (net9.0) and jf12 (net10.0) builds clean; translation validation passes (no new strings); `mkdocs build --strict` passes.
- Three adversarial review rounds (multi-agent + independent model); all confirmed findings fixed — largely in-flight-response races where a fetch spanning the switch could repopulate a freshly cleared cache.

### Notes

- No new settings; no i18n changes. Docs: `project-structure.md` documents the new core module.
- Known limitation: per-user *feature toggles* that gate non-idempotent initializers (people tags, user review tags) still need a reload to newly *enable* for the incoming user; the four base tag renderers re-initialize automatically. Data correctness (no cross-user leakage) is complete regardless.
- On upgrade, the tag/people caches are wiped once when the ownership sentinel is first stamped (they rebuild automatically).

---
*This PR was developed with AI assistance (Claude Code); the implementation has been reviewed, tested end-to-end on both supported Jellyfin versions, and I can explain any part of it.*

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
@@ -3361,6 +3361,7 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
 
         [HttpGet("user-settings/{userId}/settings.json")]
         [Authorize]
+        [ResponseCache(NoStore = true, Location = ResponseCacheLocation.None)]
         public IActionResult GetUserSettingsSettings(string userId)
         {
             var authorizationResult = AuthorizeUserConfigAccess(userId, out var authorizedUserId);
@@ -3439,6 +3440,7 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
 
         [HttpGet("user-settings/{userId}/shortcuts.json")]
         [Authorize]
+        [ResponseCache(NoStore = true, Location = ResponseCacheLocation.None)]
         public IActionResult GetUserSettingsShortcuts(string userId)
         {
             var authorizationResult = AuthorizeUserConfigAccess(userId, out var authorizedUserId);
@@ -3453,6 +3455,7 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
 
         [HttpGet("user-settings/{userId}/elsewhere.json")]
         [Authorize]
+        [ResponseCache(NoStore = true, Location = ResponseCacheLocation.None)]
         public IActionResult GetUserSettingsElsewhere(string userId)
         {
             var authorizationResult = AuthorizeUserConfigAccess(userId, out var authorizedUserId);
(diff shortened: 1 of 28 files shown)


n00bcodr:

PR #742, opened 2026-08-06
Player shortcuts without opening panels (API-driven track cycling, aspect ratio, playback info, segment skip)

4eh5xitv6787h645ebv:
# Player shortcuts without opening panels

## Description

Makes the player shortcuts act **without opening any native panels or menus**:

- **Cycle Audio / Subtitle Tracks** — switches the track through Jellyfin's own remote-control API (`POST /Sessions/{id}/Command` with `SetAudioStreamIndex` / `SetSubtitleStreamIndex`) against the plugin's own session, so no action sheet ever mounts. Verified against the jellyfin-web sources for both 10.11 and 12: `serverNotifications.js` routes both commands straight into the internal `playbackManager`, with `Index: -1` meaning subtitles off. Subtitles cycle through *Off*; a toast names the new track (escaped at the sink). The hardened OSD menu cycle from the track-cycling fix stays in place as the fallback for an unmoved surface when the session can't be resolved or a command fails.
  - The session is found via `/Sessions?ControllableByUserId=` filtered to a unique `DeviceId` match, failing open on ambiguity — this also means it works for **non-admin users** controlling their own playback (verified).
  - Robustness: presses are serialized per kind; a short-lived commanded-index memory (scoped to session + item + MediaSourceId, retired when PlayState acknowledges, cleared before any menu fallback) bridges PlayState lag so rapid presses advance through distinct tracks; item-identity guards (source-URL parse, page-URL fallback, or a press-time session probe for id-less hls.js blob sources) swallow presses whose item changed mid-flight rather than commanding the wrong item.
- **Cycle Aspect Ratio** — writes the same `appSettings` localStorage key (`aspectRatio`) the native player reads and applies `object-fit` to the video element directly (PGS subtitle canvas mapped like the native player: `auto` → `contain`), mirroring `htmlVideoPlayer.setAspectRatio`. The native settings menu's check marks stay in sync. No settings panel.
- **Show Playback Info** — toggles a plugin-rendered overlay (resolution, speed, dropped frames, buffer, play method with transcode completion %, transcoding container/codecs, bitrate, transcode reason, active audio/subtitle track names) built from the media element plus the own-session probe, instead of clicking settings → stats. Built entirely with `textContent`/`replaceChildren` (no HTML sink); one 1 s refresh timer exists only while the overlay is open, with coherence guards so a mid-probe episode change never renders mixed data, and it self-destructs when the player goes away.
- **Skip Intro/Outro** — when the server has Media Segments for the item (`GET /MediaSegments/{itemId}`, available on 10.11 and 12), seeks straight past the segment containing the current position (transcode clock offset handled); the visible skip-button click remains the fallback for items without segment data and third-party skip plugins.

16 new locale keys added to all 29 locales with real translations (`validate-translations.js` passes; upstream file formatting untouched — additive keys only). Docs updated in `enhanced/enhanced-features.md`.

## Notes / limitations

- Switching audio still restarts the stream server-side — inherent to Jellyfin track selection, identical to what the menu does.
- Track commands round-trip through the server (self-remote-control) and are delivered over the client's websocket.
- `JE.openSettings` no longer has callers after this change; left in place as an existing public surface.

## Testing

- [x] Feature works as expected — Playwright E2E on clean **Jellyfin 10.11.11** and **12.0-rc3** servers with a multi-track test item (2 audio, 2 subtitle streams), from both an **admin and a regular user**: audio index provably changes with **zero action sheets rendered**; subtitles cycle `stream → Off → stream`; aspect ratio round-trips `auto → cover → fill → auto` in both localStorage and the element style; the overlay renders live transcode data and toggles cleanly; skip falls back gracefully with no segment data.
- [x] No console errors
- [x] Compatible with Jellyfin 12 — tested on **12.0-rc3**

## AI assistance

This contribution was developed with AI assistance (implementation and an adversarial multi-round review that verified the remote-command contracts directly against the jellyfin-web 10.11 and 12 sources, and added sink-escaping for server-provided track names on both the new and pre-existing toast paths).


n00bcodr:

Issue #690, opened 2026-07-03
🐛[BUG] Auto-Skip ignores IntroSkipper offsets


Pingger:
**Describe the bug**
When Auto Intro Skip or Outro Skip are enabled, Jellyfin Enhanced ignores the configured offsets of Intro Skipper. If the IntroSkipper offset is set to 10 seconds, the "Skip Intro" Button appears 10 seconds later, but Jellyfin Enhanced does not wait those 10 seconds and thus tends to skip non-intro sections just before the intro. (similarly for outro)

**Expected behavior**
Jellyfin Enhanced should abide by the Skip Intro offset settings

**Jellyfin Environment (please complete the following information):**
 - Server Version: 10.11.11
 - Plugin Version: 11.12.0.0
 - Jellyfin Installation Method: Linux (package repository "https://repo.jellyfin.org/ubuntu")
 - Device: Jellyfin Desktop on Garuda Linux, Jellyfin Desktop on Windows 11 
 - External Access Method: direct over public ip. as well as through wireguard site-to-site VPN (just to make sure)
 - Theme being used: default (no extra themes installed or selected)
 - [x] [FileTransformation](https://github.com/IAmParadox27/jellyfin-plugin-file-transformation) Installed (2.5.11.0)

**Logs**

No Browser Console Logs, as Auto-Skip is apparently "not planned" for Firefox #499

[JellyfinEnhanced_2026-07-03.log](https://github.com/user-attachments/files/29636769/JellyfinEnhanced_2026-07-03.log)

**Additional context**

PR #744, opened 2026-08-06
feat(player): auto-skip intros/outros from Jellyfin's media segments

### LINKED ISSUE #499: Auto intro skip not working on Firefox and possibly all non-Chromium browsers
Auto intro skip does not function on Firefox. The "Intro Skipped" notification appears, indicating the plugin detects the intro and triggers the skip, but the video does not actually seek past the intro. Manually clicking the skip button works fine on all browsers.
Tested on:

Firefox (Windows) — skip notification shows but no seek
Chrome (Windows) — works correctly

Steam Deck Firefox — same behavior as windows
Steam Deck Chrome —  works correctly

All clients have identical playback settings. Server is running latest Jellyfin version in Docker on Ubuntu server with the intro-skipper and Jellyfin Enhanced plugins installed.

### LINKED ISSUE #716: 🐛[BUG] Auto-Skip doesn't trigger during media segment.
**Describe the bug**
Media segments don't seem to be automatically skipped. I see the skip button (from Intro Skipper) and nothing happens, until I press the button manually.

- Auto-Skip functionality enabled server-sided, and also client-sided in the Jellyfin Enhanced settings.
- Intro didn't skip even when Jellyfin accessed through LAN.

**Expected behavior**
Automatic skip of intro/outro/media segment.

 - Server Version: 10.11.11
 - Plugin Version: 11.12.0.0
 - Intro Skipper Version: 1.10.11.22+b5f307c978fd
 - Device: Web - Google Chrome
 - External Access Method: Cloudflare domain access without proxying.
 - Theme being used: abyss-jellyfin
 - [✅] [FileTransformation](https://github.com/IAmParadox27/jellyfin-plugin-file-transformation) Installed

**Logs**

I couldn't find any logs neither in Jellyfin's log, nor the plugin's log.

4eh5xitv6787h645ebv:
## Description

Rewrites auto-skip to read Jellyfin's native media segments (`GET /MediaSegments/{itemId}`, available on 10.10+ and 12) and seek to the segment's exact `EndTicks`, replacing the `MutationObserver` that watched for the native skip button and clicked it when its text matched `"Skip Intro"` / `"Skip Outro"`.

Closes #716
Closes #499

### Why

The old approach inherited four problems from matching on the button:

1. **Dead on localized clients.** `buttonText.includes('Skip Intro')` never matches on a German, French or Hebrew client, so auto-skip silently did nothing for non-English users.
2. **Required the button to exist.** It only worked when the user's media-segment action was *Ask to skip*. With the action set to **None** — reasonable for someone who wants the plugin to handle it — no button ever renders and nothing was skipped.
3. **Ignored the real segment boundary.** With no access to `StartTicks`/`EndTicks` it landed wherever the button seeked.
4. **Re-fired after a manual seek.** There was no seek-back guard, so seeking back into an intro made the client re-prompt and the observer clicked again, yanking the user forward. `JE.state.skipToastShown` suppressed only the *toast*, not the click.

### What it does now

`js/enhanced/player/auto-skip.js` fetches the item's media segments and, driven by the media element's `timeupdate` (no polling, bounded per-tick scan), seeks to the segment's exact `EndTicks`. It mirrors Jellyfin's own `MediaSegmentManager` semantics:

- **Backward-entry guard** — if the previous evaluated position was already past the segment start, the segment is ignored (you seeked back into it, or something else already skipped it).
- **Last-ignored latch** keyed on segment identity, so an ignored segment stays inert while it is the last-touched one — but a genuine replay (rewind to before the segment, play forward through it) skips again, exactly like the native Skip action. There is deliberately no permanent per-item "already skipped" set.
- **Sub-second segments are never skipped** (native parity, avoids stream churn).
- **Absolute-position math.** The element clock restarts at 0 on a non-`CopyTimestamps` progressive transcode, so the offset is derived from the source URL and comparisons run in absolute ticks. `.m3u8` sources are excluded — their `StartTimeTicks` is a seek hint on an absolute timeline, not a clock offset.
- **Item identity** resolves from `/Videos/{id}` in `currentSrc`, falling back to the video-page URL id, then to a now-playing session probe for `blob:` (hls.js) sources, which carry no id. The probe is single-flight and fails closed when more than one session matches the device — no auto-skip beats a skip on the wrong item. An unresolved probe is retried on later ticks (bounded, reset per source), because the first tick of a new source can beat the client's own playback-start report.
- **Fenced to `JE.session`.** The engine captures the identity epoch, drops stale work, and tears down via `onUserChange`, so an SPA user switch can't let one account's segments or in-flight probe act on another's playback.
- Stale fetches are dropped by token; a fetch failure disables auto-skip for that item view only and is never cached.

Precedence vs native is unchanged: the plugin seeks to the same `EndTicks` the native client would, so any overlap is idempotent — one jump to the identical position. Recap/Preview/Commercial stay with Jellyfin's per-type actions; the two toggles deliberately don't cover them.

### Scope

- No new settings — the existing `autoSkipIntro` / `autoSkipOutro` per-user toggles and their admin defaults are unchanged.
- No new translation keys — reuses `toast_auto_skipped_intro` / `toast_auto_skipped_outro`.
- Removes `JE.state.skipToastShown`, which existed only for the old observer.
- `playback.js` exports `getVideo` / `getCurrentVideoItemId` via `JE.internals.player` so the new module reuses them rather than duplicating the lookups.
- Docs updated: `enhanced-features.md`, `faq-support/faq.md`, `advanced/project-structure.md`.

## Implementation Notes

This PR was developed with AI assistance (Claude). I have reviewed and tested all changes and understand the implementation.

## Testing

- [x] Tested on Jellyfin 10.11 (10.11.10)
- [x] Tested on Jellyfin 12 (12.0.0)
- [x] Verified no basic errors

Verified on throwaway containers with the plugin injected normally (File Transformation on 10.11, self-injection on 12) and real H.264/AAC media playing, as admin and as a non-admin user.

**A/B against `main`**, same server, same segment, with segments visible to the client so the real Skip button renders:

| Scenario | `main` | This branch |
|---|---|---|
| German client, intro 2–25s | Button reads "Vorspann überspringen" — **no skip, no toast** | Seeks 1.82s → **25.000s**, toast "Intro automatisch übersprungen" |
| English client, intro 2–25s | Seeks 2.09s → 25.000s | Seeks 1.81s → 25.000s |
| English client, intro 20–45s | Seeks 20.15s → 45.000s | Seeks 19.88s → 45.000s |

The first row is #716 — the skip button is right there and nothing happens. The other two show no regression where the old path did work.

**Firefox 146** (the browser in #499, where the old code showed the toast but never seeked): seeks 2.22s → 25.000s, seek-back inert.

**With a real Intro Skipper install** (1.10.11.19) and fingerprint-detected segments on purpose-built media:

| Intro Skipper config | Stored segment | `main` | This branch |
|---|---|---|---|
| offsets 0 | 5.697 – 38.023s | 5.81 → 38.02 | 5.54 → 38.02 |
| `IntroStartOffset=10`, re-analyzed | 15.697 – 38.023s | 15.91 → 38.02 | 15.63 → 38.02 |

Per runtime, also verified: exactly one seek from inside the intro to the segment's exact end (asserted against a seek trace, not a `currentTime` threshold); `/MediaSegments/{itemId}` requested for the item actually playing; exactly one localized toast; seeking back into the just-skipped segment stays put; turning the setting off stops skipping; no failing plugin-served requests and no plugin console errors.

Engine logic additionally covered by a 42-case Node harness (boundary math, type gating, backward-entry guard and latch, sub-second guard, item-change resets, transcode-offset truth table, late/stale/failed fetches, session-resolver single-flight and retry bounds). The harness was mutation-tested — 17 seeded defects, each caught. It isn't included in this PR since the repo has no JS test setup.

`dotnet build` green for both `jf10` (net9.0) and `jf12` (net10.0); `node scripts/validate-translations.js validate` passes; `mkdocs build --strict` passes.

## Notes on #690

I could not reproduce it, and I don't think this PR fixes it — leaving that issue open.

Intro Skipper applies `IntroStartOffset` / `IntroEndOffset` **server-side at analysis time** (`TimeAdjustmentHelper.cs`) and persists the adjusted values, so both the old and new implementations honour them (table above). Two things that may explain the report, neither plugin-side:

1. **The offset does nothing until segments are re-analyzed.** Changing the setting — even with a full server restart — leaves existing segments untouched; only re-running *Detect and Analyze Media Segments* rewrites them. Verified: setting `IntroStartOffset=10` moved `StartTicks` by exactly +10.000s, but only after re-analysis.
2. **Chromaprint places the intro start early.** On the test media the true intro began at 8.011s and detection stored 5.697s, so auto-skip fires slightly *before* the intro actually starts — which matches "skips non-intro sections just before the intro". That's a detection artifact, identical with or without this plugin.

## Known limitations

- Segments are stored per item, not per version, so an item with multiple versions whose timings differ sees the primary version's boundaries. Documented.
- Segments with no `EndTicks` are left to the native client (which advances to the next item) — a plugin toggle auto-playing the next episode seemed too invasive.
- The seek is a plain `video.currentTime` assignment, since `playbackManager` isn't reachable from a plugin. Equivalent to the native seek for direct play/stream and HLS. On a *progressive* transcode the element can only seek within what the server has produced, so a skip there may stall or no-op where native would relaunch the stream; and the seek is local, so SyncPlay groups don't follow it. The old button-clicking path routed through native handlers in those two cases.
- Enabling a toggle *while already inside* a segment takes effect from the next segment (native reads its action map once at playback start, so this is no worse).
- The settings-panel hint still reads "Depends on Intro-Skipper Plugin being available" — accurate for the common case, and rewording it would churn all 26 locales. The docs explain that any media-segment provider works.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/config.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/config.js
@@ -23,7 +23,6 @@
         activeShortcuts: {},
         // { itemId, surface: 'continuewatching'|'nextup'|null, ts } captured on a menu trigger.
         removeContext: null,
-        skipToastShown: false,
         pauseScreenClickTimer: null
     };
 
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/player/auto-skip.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/enhanced/player/auto-skip.js
@@ -1,0 +1,556 @@
+/**
+ * @file Auto-Skip — data-driven segment auto-skip that honours Jellyfin's native
+ * Media Segment boundaries exactly.
+ *
+ * The previous implementation auto-CLICKED the native "skip" button by matching
+ * its English text ("Skip Intro"/"Skip Outro"). That inherited three bugs:
+ *   • dead on localized clients (the button text is translated);
+ *   • only worked when the native client was configured to PROMPT — a user whose
+ *     media-segment action is "None" never gets a button, so never got a skip;
+ *   • no seek-back guard — the click re-fired every time the native client
+ *     re-prompted after a manual seek, insta-re-skipping the user.
+ * It also never read the segment's StartTicks/EndTicks, so it could not honour
+ * the real boundary; it landed wherever the native button happened to seek.
+ *
+ * This engine reads the native Media Segments (GET /MediaSegments/{itemId},
+ * available on Jellyfin 10.10+ and 12) and seeks to the segment's exact EndTicks
+ * itself, driven by the media element's `timeupdate` event (no polling; the
+ * per-tick scan is bounded by the handful of segments an item has). It mirrors
+ * the native MediaSegmentManager's guards (backward-entry `lastTime > StartTicks`
+ * plus a last-ignored latch — deliberately NO permanent once-per-item set, so a
+ * legitimate replay skips again exactly like native), meaning a user who seeks
+ * back into a segment after an auto-skip is never insta-re-skipped — including
+ * when the native client's own Skip action performed the original skip.
+ *
+ * Precedence vs native: the plugin seeks to the SAME EndTicks the native client
+ * would, so any overlap with the native Skip action is idempotent (one visible
+ * jump to the identical position). The plugin's value-add is the in-player quick
+ * toggle (Auto Skip Intro / Auto Skip Outro) plus a localized toast; the native
+ * per-type actions still own Recap/Preview/Commercial, which these toggles
+ * deliberately do not cover.
+ *
+ * One deliberate divergence from the native manager: a segment with no EndTicks is
+ * ignored rather than acted on. Native reads "no end" as "advance to the next
+ * track", which is far too invasive for a plugin toggle to do on the user's
+ * behalf; the native per-type action still handles that case if the user wants it.
+ *
+ * The seek is a plain `video.currentTime` assignment, because playbackManager is
+ * not reachable from a plugin. That is equivalent to the native seek for direct
+ * play/stream and HLS — the whole item is seekable. On a *progressive* transcode
+ * the element can only seek within what the server has already produced, so a
+ * skip there may stall or no-op where the native client would relaunch the stream
+ * at the target; the toggles are best paired with direct play or HLS transcoding
+ * (jellyfin-web's default in browsers). For the same reason the seek is local
+ * only, so SyncPlay groups will not follow it.
+ */
+(function(JE) {
+    'use strict';
+
+    /** 1 tick = 100 ns; 10,000,000 ticks per second. */
+    const TICKS_PER_SECOND = 10000000;
+
+    /** Mirror the native client: never skip a segment shorter than this (avoids churn). */
+    const MIN_SEGMENT_TICKS = TICKS_PER_SECOND; // 1 second
+
+    /**
+     * How many times a single media source may be probed for its item id before
+     * we give up on it. Ticks drive the retries (~4/s), so this is roughly a
+     * second of grace for the playback-start report to land.
+     */
+    const MAX_PROBE_ATTEMPTS = 5;
+
+    /**
+     * @typedef {Object} MediaSegment
+     * @property {string} [Id]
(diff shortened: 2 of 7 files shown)


n00bcodr:

PR #748, opened 2026-08-07
Add official German FSK colors and normalization

e1ns2wo:
This PR updates German FSK rating handling in Jellyfin Enhanced.

Changes:
- Normalize German rating variants such as DE-12, FSK12 and FSK 12 to FSK-12
- Display German ratings consistently as FSK-0, FSK-6, FSK-12, FSK-16 and FSK-18
- Apply the official German FSK color scheme:
  - FSK-0: white
  - FSK-6: yellow
  - FSK-12: green
  - FSK-16: blue
  - FSK-18: red
- Keep non-German rating display text unchanged
- Update the ratings documentation screenshot accordingly

PR #749, opened 2026-08-09
Fix German FSK rating colors

e1ns2wo:
Follow-up to #748.

The German FSK normalization was merged successfully, but the dedicated FSK color overrides were not included in the final merge commit.

This restores the intended official German FSK color scheme:

- FSK-0: white
- FSK-6: yellow
- FSK-12: green
- FSK-16: blue
- FSK-18: red

All FSK labels use black text for readability and to match the official FSK design.

### DIFF
--- a/css/ratings.css
+++ b/css/ratings.css
@@ -444,6 +444,48 @@
     border-color: #C2185B !important;
 }
 
+/* === GERMANY: Official FSK color scheme === */
+
+/* FSK 0 — White */
+.mediaInfoOfficialRating[rating='FSK-0'] {
+    background-color: #FFFFFF !important;
+    border-color: #BDBDBD !important;
+    color: #000000 !important;
+    text-shadow: none !important;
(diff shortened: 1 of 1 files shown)


n00bcodr:
The ratings probably didnt reflect due to the jsdelivr caching, I just purged the cache now. 

But I think why #748 used black now makes sense looking at https://www.fsk.de/english/#ratings_advices

Although it is out of the already existing pattern, it is the right styling for the rating, proceeding to merge this and will trigger the purge cache on jsdelivr again, but you would have to run the "Refresh CDN Assets" scheduled task for these changes to appear.

e1ns2wo:
It looks good now. Thanks a lot! :)
Yes, black was actually intentional because it matches the official schema. I noticed that you usually use white everywhere else, but since FSK-0, for example, has a white background, using black text makes more sense.

Best regards :)

PR #761, opened 2026-08-15
fix: honour "Remove from Continue Watching" on Home Screen Sections rows

### LINKED ISSUE #750: 🐛[BUG] Element removed from continue watching reappear after refresh
**Describe the bug**  
After removing an item from Continue Watching, it disappears as expected, but it reappears after refreshing/reloading the page.

The issue occurs across multiple clients.

**Expected behavior**  
The removed item should remain removed from Continue Watching after a refresh.

**Jellyfin Environment:**  
- Server Version: 10.11.1
- Plugin Version: 12.1
- Device:
  - Jellyfin Android App
  - Jellyfin Web on Firefox
  - Jellyfin Web on Chrome
  - Windows Desktop App
  - Linux Desktop App
- Theme being used: Elegantfin

**Additional information**  
Normal Jellyfin playback-state functions work correctly, including:
- Mark as played
- Mark as unplayed
- Items leaving Continue Watching normally after playback is completed

The issue appears to affect specifically the **Remove from Continue Watching** function: the item is removed immediately from the UI, but the removal does not persist after a refresh.

4eh5xitv6787h645ebv:
## Description

Fixes "Remove from Continue Watching" not sticking: the item disappears, but comes back on the next refresh.

Closes #750

There turned out to be two independent reasons an item can come back, so this is two commits. The first is
what the reporter hit; the second is why nothing caught it.

### 1. Home Screen Sections rows were never filtered

Remove-from-Continue-Watching is enforced **server-side**: the hide is written to `hidden-content.json` and
`HiddenContentResponseFilter` strips the item from the list endpoints that render the row. That filter's route
table only knew about Jellyfin's own controllers (`Items.GetResumeItems`, `TvShows.GetNextUp`, …).

The [Home Screen Sections](https://github.com/IAmParadox27/jellyfin-plugin-home-sections) plugin doesn't use
those endpoints — it builds **every** home row from its own `GET /HomeScreen/Section/{sectionType}`. That
endpoint was never filtered, so on an HSS home screen the card vanished optimistically, the hide was written
correctly, and then it came straight back.

Reproduced on a clean 10.11.11 container (JE 12.2.0.0 + File Transformation 2.5.11.0 + HSS 2.5.11.0):

```
after "Remove from Continue Watching":
  /UserItems/Resume                    -> Alpha Movie, Test Show S01E01           (filtered, correct)
  /HomeScreen/Section/ContinueWatching -> Beta Movie, Alpha Movie, Test Show ...  (unfiltered, item is back)
```

Two details made this more than a one-line table entry:

* **The surface isn't fixed per route** — it depends on which section was requested. That entry's surfaces are
  `null` and `ResolveSurfaces` derives them per request from the `{sectionType}` route value:
  `ContinueWatching` → `continuewatching`, `NextUp` → `nextup`, everything else → `library`, so unknown and
  future section types still get global hides applied and no JE update is needed when HSS adds one.
* **HSS has a combined `ContinueWatchingNextUp` row** that renders resume items and next-up items in one
  list. Both scopes are in play there, but they are applied *per item* rather than to the whole list: only a
  resume item carries a playback position, so that is what decides which surface each item is appearing as.
  Applying the union to the entire row would mean a Next-Up hide also suppressed that episode once the user
  started watching it and it became a Continue Watching entry — invisible everywhere else, since the native
  resume row and HSS's separate Continue Watching row would both still show it. This matches what the
  client-side detector does per card, so the two halves agree.

Matching is by controller/action name only, so there is **no compile-time or runtime dependency on the
plugin**: with HSS absent the route never matches and nothing runs. Behaviour of every existing native route
is unchanged, including the `/Items` + `searchTerm` → `search` special case and the metadata-resolver `Ids`
bypass.

### 2. The client-side filter could never match anything

JE also filters cards in the DOM, which is what should have made the gap above invisible to users. It was
broken in two ways, so it had been doing nothing for scoped hides:

* **ID format.** The server writes GUIDs dashed into `hidden-content.json`; card `data-id` attributes are
  undashed. They were compared raw, so a server-written entry never matched a card. A `normalizeId` helper now
  sits on every store/DOM comparison. The same mismatch also made `unhideItem`'s fallback lookup miss, so
  un-hiding such an entry from a details page or a card button silently did nothing.
* **Row identification by heading text.** A card's row was identified by matching the English strings
  `"continue watching"` / `"next up"` in the heading, so on a translated client nothing was ever classified.
  A native row is now read from the user's own `homesection{N}` settings — the same configuration
  jellyfin-web itself builds the home screen from, and which states outright what each `sectionN` slot is —
  so the answer is exact rather than inferred, in any language. That config is fetched once per user in the
  background; until it arrives, and for rows it cannot describe (HSS rows, and the per-library sub-rows
  inside the Latest Media slot), detection falls back to markup: HSS's section-id class, jellyfin-web's Next
  Up link, and the `data-monitor` marker it sets only on its resume and next-up rows, with the heading last.
  A row holding both kinds at once is resolved per card by its playback position.

Detection is deliberately **split in two**, because the two callers are asking different questions.
Filtering and the library hide button need the strict row test: a resume item also appears in rows like
Recently Added and those cards carry a playback position too, so treating the position alone as proof of
surface would blank the item wherever it appears. Choosing the scope to *write* for a Remove action stays
generous and keeps its previous behaviour.

## Implementation Notes

This PR was developed with AI assistance (Claude). I have reviewed and tested all changes and understand the
implementation.

## Testing

Verified on a purpose-built Jellyfin **10.11.11** container (JE 12.2.0.0, File Transformation 2.5.11.0, Home
Screen Sections 2.5.11.0) with a generated library, driven through the real web UI with Playwright as both an
admin and a non-admin account. Native-surface regressions were re-checked separately against a 10.11 server
with a real library.

- [x] Verified no basic errors

**The reported bug**

- [x] Remove from Continue Watching on an HSS home screen → item stays gone after a full page reload
- [x] Same item also gone from HSS's combined *Continue Watching / Next Up* row
- [x] `/HomeScreen/Section/ContinueWatching` now matches `/UserItems/Resume` exactly
- [x] `/HomeScreen/Section/NextUp` honours a Next-Up-scoped hide; native `/Shows/NextUp` agrees
- [x] Combined row drops both a CW-scoped and a Next-Up-scoped hide, keeps everything else
- [x] Combined row resolves scope per item: an episode hidden from Next Up is dropped while it *is* a Next Up
      item, and comes back once it has a playback position and belongs in Continue Watching — matching the
      native resume row and HSS's separate Continue Watching row
- [x] An action that throws (a plugin section doing network I/O to a service that is down) is left alone
      rather than being reported as a response-shape change

**No over-hiding**

- [x] A Continue Watching hide does **not** remove the item from `/Items`, Recently Added, or any library row
- [x] Unmapped HSS section (`RecentlyAddedMovies`) drops a *global* hide but not a scoped one
- [x] Per-user isolation through HSS's section cache: two users with different hides each see only their own

**Client-side filter**

- [x] With the server-side half deliberately absent, the DOM filter alone keeps the item hidden after reload —
      i.e. it now works as the backstop it was meant to be
- [x] On an Italian client: "Continua a guardare" → `continuewatching`, combined row → per card,
      "Film aggiunti di recente" → `null`, native Next Up row → `nextup`
- [x] Row identity follows the user's configuration, not the markup: with Continue Watching moved to a
      different `homesection` slot it is still resolved correctly, and a row relabelled into a slot the user
      configured as Next Up resolves as `nextup` even though its markup and card playback positions say
      resume — i.e. the configuration is authoritative where the two disagree
- [x] Falls back cleanly when the configuration has not loaded yet or cannot describe a row
- [x] Hide/unhide round-trips with both dashed and undashed IDs, including removing the store entry
- [x] Library hide button still writes a *global* hide outside a Continue Watching / Next Up row

**Build**

- [x] Compiles clean for both targets (net9.0 / 10.11 and net10.0 / 12), no new warnings
- [x] `validate-translations.js validate` passes (no new keys)
- [x] `mkdocs build --strict` passes

No new settings, no new translation keys, no UI changes — so no screenshots. Docs updated in
`enhanced-features.md` and `faq.md` to say the removal is honoured on HSS home screens.

## Not included

While tracking this down I found a third, unrelated way to get the same symptom, with no other plugin
involved: the "Remove from Continue Watching" toggle in the user settings panel is per-user, but the
server-side filter is gated on the admin-level `RemoveContinueWatchingEnabled` / `HiddenContentEnabled`
config, both of which default to `false`. A user who enables the toggle themselves on a server where the admin
never did gets the menu item, a successful hide, and no filtering — so the item returns on every client, and
the workaround is for the admin to enable it in the plugin config.

That looked like a deliberate design decision rather than a bug, so I've left it alone. Happy to open it as a
separate issue or PR if you'd like it changed.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/PluginServiceRegistrator.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/PluginServiceRegistrator.cs
@@ -68,8 +68,9 @@ namespace Jellyfin.Plugin.JellyfinEnhanced
             serviceCollection.AddTransient<ClearTranslationCacheTask>();
 
             // Hidden Content: server-side filter for every native Jellyfin endpoint that surfaces user-facing item lists
-            // (Resume, Items, Latest, NextUp, Upcoming, Suggestions, SearchHints). Same filter handles "Remove from
-            // Continue Watching" via HideScope=continuewatching in hidden-content.json.
+            // (Resume, Items, Latest, NextUp, Upcoming, Suggestions, SearchHints), plus the Home Screen Sections
+            // plugin's own section-content endpoint, which replaces those native lists on an HSS home screen. Same
+            // filter handles "Remove from Continue Watching" via HideScope=continuewatching in hidden-content.json.
             serviceCollection.AddSingleton<MaintenanceModeService>();
             serviceCollection.AddSingleton<HiddenContentResponseFilter>();
             serviceCollection.AddScoped<IEventConsumer<PlaybackStartEventArgs>, ContinueWatchingPlaybackConsumer>();
--- a/Jellyfin.Plugin.JellyfinEnhanced/Services/HiddenContentResponseFilter.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Services/HiddenContentResponseFilter.cs
@@ -35,20 +35,67 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Services
                 _hcCache.TryRemove(userId, out _);
         }
 
-        private static readonly Dictionary<(string, string), (string Surface, ResponseHandler Handler)> _routes
+        // Shared immutable surface sets. Kept as statics so neither route matching nor per-request
+        // surface resolution allocates.
+        private static readonly string[] ContinueWatchingSurface = { "continuewatching" };
+        private static readonly string[] LibrarySurface = { "library" };
+        private static readonly string[] NextUpSurface = { "nextup" };
+        private static readonly string[] UpcomingSurface = { "upcoming" };
+        private static readonly string[] RecommendationsSurface = { "recommendations" };
+        private static readonly string[] SearchSurface = { "search" };
+
+        // A route whose Surfaces is null resolves its surface(s) per request - see ResolveSurfaces.
+        private static readonly Dictionary<(string, string), (string[]? Surfaces, ResponseHandler Handler)> _routes
             = new(KeyComparer.Instance)
         {
-            { ("Items", "GetResumeItems"),         ("continuewatching", FilterQueryResult) },
-            { ("Items", "GetResumeItemsLegacy"),   ("continuewatching", FilterQueryResult) },
-            { ("Items", "GetItems"),               ("library",          FilterQueryResult) },
-            { ("Items", "GetItemsByUserIdLegacy"), ("library",          FilterQueryResult) },
-            { ("UserLibrary", "GetLatestMedia"),       ("library", FilterEnumerable) },
-            { ("UserLibrary", "GetLatestMediaLegacy"), ("library", FilterEnumerable) },
-            { ("TvShows", "GetNextUp"),                ("nextup",          FilterQueryResult) },
-            { ("TvShows", "GetUpcomingEpisodes"),      ("upcoming",        FilterQueryResult) },
-            { ("Suggestions", "GetSuggestions"),       ("recommendations", FilterQueryResult) },
-            { ("Suggestions", "GetSuggestionsLegacy"), ("recommendations", FilterQueryResult) },
-            { ("Search", "GetSearchHints"),            ("search",          FilterSearchHints) },
+            { ("Items", "GetResumeItems"),         (ContinueWatchingSurface, FilterQueryResult) },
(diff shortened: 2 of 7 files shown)


n00bcodr:
Available in version [12.3.0.0](https://github.com/n00bcodr/Jellyfin-Enhanced/releases/tag/12.3.0.0)

PR #790, opened 2026-08-28
fix(tags): preserve Matroska LanguageBCP47 regions

### LINKED ISSUE #789: 🐛[BUG] Language Tags lose Matroska LanguageIETF/BCP-47 regional variants before flag resolution
## Describe the bug

PR #739 added region-aware language flag resolution (`pt-BR` → Brazil, `en-US` → US, etc.), but with normal Matroska files produced by MKVToolNix the regional BCP-47 value does not appear to reach Jellyfin Enhanced.

Matroska can store both:

- the legacy `Language` element, using an ISO 639 code such as `por` or `eng`
- the newer `LanguageIETF` / BCP-47 value, such as `pt-BR`, `pt-PT`, `en-US` or `en-GB`

`mkvmerge -J` correctly exposes both values.

However, both regular `ffprobe` and Jellyfin's bundled `ffprobe` expose only the legacy language value. Jellyfin Enhanced then stores that same base value in `AudioLanguages`.

As a result, the region-aware resolver added in #739 never receives the regional tag for these files.

In other words, the flag resolver appears to work correctly when it receives `pt-BR`, `en-US`, etc., but with standard Matroska `Language` + `LanguageIETF` metadata it receives only `por`, `eng`, etc.

This may also explain the report in #739 from a user who installed the released fix but still only saw the Portuguese flag.

---

## Expected behavior

When a Matroska audio track contains an explicit `LanguageIETF` / BCP-47 value, Jellyfin Enhanced should preserve and use it.

For example:

```text
Language=por + LanguageIETF=pt-PT → Portugal flag
Language=por + LanguageIETF=pt-BR → Brazil flag

Language=eng + LanguageIETF=en-US → United States flag
Language=eng + LanguageIETF=en-GB → United Kingdom flag
```

If there is no regional BCP-47 value, the existing default behavior should remain unchanged:

```text
por → Portugal
eng → United Kingdom
spa → Spain
fra → France
etc.
```

A regional variant should only be used when the media explicitly provides one. It should not be guessed from the base language.

---

## Reproduction / diagnostics

I tested four real MKV files containing explicit regional language metadata.

Environment:

- Jellyfin Server: `12.0.0`
- Jellyfin Enhanced: `12.4.1.0`
- Installation: Docker
- Jellyfin ffprobe: `8.1.2-Jellyfin`
- Server-Side Tag Cache: enabled

The results were:

| Matroska metadata (`mkvmerge -J`) | Host ffprobe | Jellyfin ffprobe | Enhanced `AudioLanguages` |
|---|---|---|---|
| `por / pt-PT` | `por` | `por` | `por` |
| `por / pt-BR` | `por` | `por` | `por` |
| `eng / en-US` | `eng` | `eng` | `eng` |
| `eng / en-GB` | `eng` | `eng` | `eng` |

### Example: Brazilian Portuguese

`mkvmerge -J`:

```text
ID=1 | lang=por | ietf=pt-BR | name=Brazilian
```

The same file with Jellyfin's ffprobe:

```json
{
  "index": 1,
  "codec_name": "eac3",
  "tags": {
    "language": "por",
    "title": "Brazilian"
  }
}
```

And the corresponding Jellyfin Enhanced server-side tag cache contains:

```json
"AudioLanguages": [
  "por",
  "eng"
]
```

So the data flow is effectively:

```text
Matroska:
Language=por
LanguageIETF=pt-BR
        ↓
FFmpeg / Jellyfin:
Language=por
        ↓
Jellyfin Enhanced:
AudioLanguages=["por"]
        ↓
media-language.js:
por
        ↓
Portugal flag
```

The same loss occurs with `pt-PT`, `en-US` and `en-GB`.

---

## Why this appears to happen

The server-side cache currently derives audio languages from Jellyfin's `MediaStream.Language`.

In `TagCacheService`, the relevant logic effectively does:

```csharp
if (s.Type == MediaStreamType.Audio && !string.IsNullOrEmpty(s.Language))
{
    var lang = s.Language.ToLowerInvariant();

    if (lang != "und" && lang != "root")
    {
        languages.Add(lang);
    }
}
```

The normal client-side Language Tags path also reads:

```javascript
stream.Language
```

So both paths depend on Jellyfin exposing the regional tag through `Language`.

The problem is that, for the tested Matroska files, FFmpeg/Jellyfin only exposes the legacy Matroska `Language` element (`por`, `eng`, etc.) and not `LanguageIETF`.

This means `media-language.js` never gets the `pt-BR`, `en-US`, etc. value that it now knows how to resolve.

---

## Relation to #544 / #739

#739 correctly added support for region-aware flag resolution and stopped regional tags from being stripped client-side.

However, its description states:

> Server-side extraction already preserved regions — the loss was purely client-side

That does not appear to be true for standard MKVToolNix Matroska files using the normal combination of:

```text
Language=por
LanguageIETF=pt-BR
```

or:

```text
Language=eng
LanguageIETF=en-US
```

On Jellyfin 12.0.0 these reach Jellyfin Enhanced only as `por` and `eng`.

#739 was runtime-tested on Jellyfin 10.11.11, while Jellyfin 12 was compile-tested but not separately runtime-tested.

There is also a follow-up comment on #739 from a user who installed version `12.2.0`, restarted Jellyfin and still reported no visible difference.

---

## Possible fix direction

I think the ideal solution would be to preserve the authoritative BCP-47 value before it reaches `media-language.js`, rather than adding more hard-coded aliases or trying to infer regions from track names.

Possible approach:

1. Prefer a full BCP-47 / IETF language value whenever Jellyfin or FFmpeg exposes one.
2. For local Matroska files, if Jellyfin only exposes the legacy ISO 639 value, obtain `LanguageIETF` / `LanguageBCP47` directly from the Matroska track metadata if feasible.
3. Fall back to `MediaStream.Language` for non-Matroska media or files without regional metadata.
4. Preserve the full BCP-47 value in both the server-side cache and the non-server-cache Language Tags path.
5. If the cached language semantics change, invalidate/rebuild older server-side tag-cache entries.

I would avoid using track-title heuristics such as `Brazilian`, `English (UK)`, `Latino`, etc. as the primary solution because the Matroska container already contains authoritative structured language metadata.

Once the full BCP-47 tag reaches `media-language.js`, the existing resolver already handles normal regional tags generically, so this should not require maintaining a huge manual `language-region → flag` mapping.

---

## Jellyfin Environment

- Server Version: `12.0.0`
- Plugin Version: `12.4.1.0`
- Jellyfin FFmpeg: `8.1.2-Jellyfin`
- Device: Jellyfin Web; the issue was reproduced at the server/tag-cache level and is therefore browser-independent
- External Access Method: not relevant; reproduced locally
- Server-Side Tag Cache: enabled

`LanguageTagsPriority` is not involved in the issue. It only affects ordering/visibility of language flags and does not change the extracted language code.

---

## Logs / diagnostic summary

There is no server error because every layer accepts the legacy language value normally.

The relevant diagnostic mismatch is the metadata itself:

```text
MKVToolNix : por / pt-BR
JF ffprobe : por
Enhanced   : por
```

and:

```text
MKVToolNix : eng / en-US
JF ffprobe : eng
Enhanced   : eng
```

maghuro:
## Summary

Preserves explicit Matroska `LanguageBCP47` / `LanguageIETF` regional variants before they are lost by Jellyfin/FFmpeg, allowing the existing region-aware language flag resolver to receive values such as `pt-BR`, `pt-PT`, `en-US`, and `en-GB`.

Fixes #789.

## Root cause

Matroska can contain both a legacy `Language` element (`por`, `eng`, etc.) and an authoritative BCP-47 language element (`pt-BR`, `en-US`, etc.).

For the tested MKVToolNix files, FFmpeg/Jellyfin exposes only the legacy/base value through `MediaStream.Language`. As a result, Jellyfin Enhanced never receives the regional tag even though its client-side flag resolver already understands it.

## Implementation

- Adds a lightweight internal Matroska/EBML reader for local Matroska/WebM sources.
- Reads only the container metadata required to locate `Tracks` / `TrackEntry`; it does not scan media clusters.
- Prefers `LanguageBCP47` over the legacy `Language` element.
- Falls back to Jellyfin's existing `MediaStream.Language` for unsupported, inaccessible, malformed, non-Matroska, or non-regional media.
- Aligns Matroska languages only with embedded audio streams and only when the embedded audio-track counts match exactly, avoiding incorrect ordinal assignment when external or missing streams are present.
- Uses the resolved language in both server-side tag-cache extraction and the `/tag-data` path used when the server cache is disabled.
- Updates item-details and Series/Season first-episode language paths to use the same region-aware projection.
- Bumps the tag-cache schema from v2 to v3 so stale entries containing region-less language values are discarded and rebuilt.
- Adds no external executable or package dependency.

Bare-language policy is intentionally unchanged. For example, an explicit `en-US` can resolve to the US flag, while bare `eng` continues to follow Jellyfin Enhanced's existing default behavior.

## Testing

Runtime-tested on:

- Jellyfin Server 12.0.0
- Jellyfin Enhanced 12.4.1.0
- Docker / Jellyfin Web
- Server-Side Tag Cache enabled and disabled

Verified with real Matroska files containing:

- `por` + `pt-PT`
- `por` + `pt-BR`
- `eng` + `en-US`
- `eng` + `en-GB`

Verified regional values in the rebuilt v3 server tag cache, including `pt-PT`, `pt-BR`, `en-US`, `en-GB`, `fr-CA`, and `es-419`.

Also verified the non-cache card path and item-details path, including Series/Season first-episode language resolution.

Build/static validation:

- Jellyfin 12 target / .NET 10: 0 warnings, 0 errors
- Jellyfin 10.11 target / .NET 9: 0 warnings, 0 errors
- `node --check` passed for both modified JavaScript files
- `git diff --check` passed

## AI assistance

AI assistance was used during investigation and implementation; I reviewed and tested the changes and understand the implementation.


n00bcodr:
Thank you! :)

PR #792, opened 2026-08-30
fix: make Seerr search React-safe and responsive

wsimon1982:
## Summary
- Keep Seerr results visually integrated with native Jellyfin search while all Enhanced-owned nodes stay under document.body.
- Remove React-subtree insertion and reposition observers that caused WebKit NotFoundError/removeChild.
- Use only the native Jellyfin search input; no duplicate field or floating control.
- Invalidate stale debounced searches synchronously and cancel pending work on navigation and refresh.
- Reserve layout space without spacer nodes inside React-owned DOM.
- Reflow after React container replacement, host resize, and iOS VisualViewport changes.
- Bound results to 8 mobile and 20 desktop.

## Verification
- Node isolation tests: 6/6 passed
- JavaScript syntax checks: passed
- git diff --check: passed
- Jellyfin 10 / .NET 9 build: 0 warnings, 0 errors
- Jellyfin 12 / .NET 10 build: 0 warnings, 0 errors
- Chromium and WebKit acceptance covered repeated native queries and absence of NotFoundError/removeChild.

## Security
- No client-side Seerr credentials.
- No private hosts, domains, keys, or tokens added.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj
+++ b/Jellyfin.Plugin.JellyfinEnhanced/JellyfinEnhanced.csproj
@@ -12,8 +12,9 @@
 
         <ImplicitUsings>enable</ImplicitUsings>
         <Nullable>enable</Nullable>
-        <AssemblyVersion>12.4.1.0</AssemblyVersion>
-        <FileVersion>12.4.1.0</FileVersion>
+        <AssemblyVersion>12.4.1.7</AssemblyVersion>
+        <FileVersion>12.4.1.7</FileVersion>
+
         <RootNamespace>Jellyfin.Plugin.JellyfinEnhanced</RootNamespace>
         <AssemblyName>Jellyfin.Plugin.JellyfinEnhanced</AssemblyName>
     </PropertyGroup>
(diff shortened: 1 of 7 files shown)


4eh5xitv6787h645ebv:
This doesn't work on jellyfin 12 r6.

4eh5xitv6787h645ebv:
Thanks for digging into this one — the crash is real and nasty. Unfortunately, testing this branch against Jellyfin 12.0-rc6 and the current unstable image shows the Seerr search results never render at all with these changes: jellyfin-web rewrites `#/search?query=…` on every keystroke, which fires the SPA-navigate event, and the new `onNavigate` handler here (`clearTimeout` + `searchGeneration++` + `closeJellyseerrSearch()`) cancels its own pending search on each keystroke. So the NotFoundError goes away, but only because the integration stops running (no `/api/v1/search` requests ever leave the browser, no cards/host are created).

The actual root cause turned out to be a single destructive DOM write — rewriting the React-owned `.noItemsMessage` textContent — and there's a minimal fix for exactly that (with the full feature set kept) in the replacement PR. Closing in favor of that one.

PR #803, opened 2026-09-06
fix(language tags): show a "no dialogue" badge for zxx instead of no flag

Kisnov:
## Problem

Audio tracks tagged `zxx` (ISO 639-2, "no linguistic content") get no flag at
all. `resolveFlag()` returns `null` for the unmapped base, so
`insertLanguageTags` never creates an `` and the card renders as if the
item had no audio language data.

That's indistinguishable from an untagged file, but the two mean opposite
things: `und` means nobody tagged it, while `zxx` is a positive statement that
there is no dialogue — silent films, and films with a score or sound design but
no speech. The distinction is worth surfacing.

## Change

Adds `zxx` to `baseLanguageFlags`, resolving to a synthetic token rather than a
flag-icons code, and renders it as a crossed-out speech bubble.

Because no country is being claimed, the badge deliberately isn't a flag.
`flags/4x3/xx.svg` ("Unknown" in flag-icons' own `country.json`) was the
alternative, but it reads as missing metadata rather than as a deliberate
"no dialogue".

Three pieces:

- **`js/core/media-language.js`**
  - `zxx` (plus its `Intl.DisplayNames` names) → the token `'zxx'`.
  - New `specialFlags` table holding flag tokens that are *not* flag-icons
    codes, as self-contained data URIs — no request, can't 404, nothing to
    add to `CdnAssetService`'s allow-list.
  - New `flagSrc(code)` helper: `specialFlags` first, `JE.cdn.flagSvg()`
    otherwise.
  - Guard in `resolveFlag` so a synthetic token can't be "refined" by a region
    or script subtag (`zxx-US` stays "no dialogue", not the US flag).
- **`js/tags/languagetags.js`** — overlay uses `flagSrc()` instead of
  `JE.cdn.flagSvg()`.
- **`js/enhanced/itemdetails/features-details-media-info.js`** — same swap for
  the details-page audio row.

The two call sites have to change together with the core: handing `'zxx'` to
`JE.cdn.flagSvg()` would build a 404 URL and `img.onerror` would remove the
element, which is where we started.

## Notes

- `flagSrc()` is a thin wrapper; `JE.cdn.flagSvg()` is unchanged and still used
  for every real flag code.
- The badge carries `data-lang="zxx"`, so it can be restyled independently via
  custom CSS.
- No cache migration needed. The overlay cache stores `{name, code}` pairs and
  `resolveFlag` runs at render time, so already-cached `zxx` items pick this up
  on the next page load.
- `und`/`root` behaviour is unchanged — still filtered in
  `extractLanguagesFromItem`, still no badge.

## Testing

Verified against the edited resolver that nothing else moved: `en`, `eng`,
`pt-BR`, `es-419`, `zh-Hant`, `ca`, `en-UK`, `pt-XA` and
`en-u-ca-gregory` all resolve as before, and unmapped/malformed codes
(`xyz`, `qq`, empty) still produce no flag. `zxx`, `ZXX` and `zxx-US` all
resolve to the new badge.


Kisnov:
icon:
zxx-no-dialogue-preview

n00bcodr:
Thank you! :)

PR #810, opened 2026-09-08
fix(custom tabs): defer tab-hosted content until its tab is active

0-zach:
I went looking for why my home page was slow and found that the Calendar tab was fetching on every home load, even though I never had it open.

`findActiveContainer()` checks `.page` before `.tabContent.is-active`:

```js
// 1. Standard Jellyfin page structure
var page = el.closest('.page');
if (page && !page.classList.contains('hide')) return el;
// 2. Custom Tabs wraps content in .tabContent.is-active (no .page ancestor)
var tabContent = el.closest('.tabContent');
if (tabContent && tabContent.classList.contains('is-active')) return el;
```

The comment on the second branch says Custom Tabs panels have no `.page` ancestor, but on 10.11 Custom Tabs injects its panels into `#indexPage`, and that isn't `.hide` while the home page is showing. So the first branch always matches and the second one never gets a chance to run — every tab-hosted module mounts and fetches on home load regardless of which tab is actually selected.

In my case that meant a `GET /arr/calendar` on every single home page load, into a tab that was closed and not even visible. That endpoint fans out to Sonarr/Radarr with no caching, so it's not a cheap thing to be firing off for nothing.

The same function is copy-pasted across six modules, so they all have it:

- `arr/calendar/calendar-custom-tab.js`
- `arr/requests/requests-custom-tab.js`
- `enhanced/hiddencontent/hidden-content-custom-tab.js`
- `extras/activity-custom-tab.js`
- `jellyseerr/recommendations/recommendations-custom-tab.js`
- `enhanced/bookmarks/bookmarks-library-render.js`

## The fix

When the element is tab-hosted, let the tab's own active state decide; otherwise fall through to the existing `.page` / visibility logic.

Nothing else needed touching, though I did check one thing first. The observer is registered with `childList`/`subtree` and no `attributes`, so if switching tabs were only a class toggle, deferring the mount would have meant it never mounted at all. It isn't — Jellyfin adds and removes nodes when it switches tabs, so the observer still fires and `tryMount()` runs on click.

## Testing

Jellyfin 10.11.7 with Custom Tabs, built with `-p:JellyfinTarget=jf10`:

On the home page there are now no `/arr/calendar` requests at all. Clicking the tab fires exactly one and the calendar renders normally. Before the change it was one request on every home load.

## Second commit

Unrelated and easy to drop if you'd rather keep this focused. The calendar loading state reuses `.je-calendar-empty` and has no `role` or `aria-live`, so screen readers get nothing at all while it loads, and on a slow *arr response the tab just reads as empty rather than busy. Gave it its own class with a spinner and a live region. Under `prefers-reduced-motion` I slowed the spinner rather than removing it, since dropping the animation entirely would leave no progress cue for the people who opt into that.

One caveat: I've only been able to test the `jf10` target. It's JS-only so the `jf12` build shouldn't care, but I can't verify that without spinning up an instance.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar/calendar-custom-tab.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar/calendar-custom-tab.js
@@ -86,14 +86,21 @@
     var all = document.querySelectorAll('.jellyfinenhanced.calendar');
     for (var i = all.length - 1; i >= 0; i--) {
       var el = all[i];
-      // 1. Standard Jellyfin page structure
+      // 1. Tab-hosted content: the tab's own active state is authoritative.
+      //    Checked BEFORE .page because Custom Tabs injects its panels into
+      //    #indexPage, which is NOT .hide while the home page is showing. A
+      //    .page-first check therefore matches every tab on load and mounts
+      //    content for tabs the user never opened.
+      var tabContent = el.closest('.tabContent');
+      if (tabContent) {
+        if (tabContent.classList.contains('is-active')) return el;
+        continue;
+      }
+      // 2. Standard Jellyfin page structure
       var page = el.closest('.page');
       if (page && !page.classList.contains('hide')) return el;
-      // 2. Custom Tabs wraps content in .tabContent.is-active (no .page ancestor)
-      var tabContent = el.closest('.tabContent');
-      if (tabContent && tabContent.classList.contains('is-active')) return el;
       // 3. Last resort: element is simply visible in the document
-      if (!page && !tabContent && el.offsetParent !== null) return el;
+      if (!page && el.offsetParent !== null) return el;
     }
     return null;
   }
--- a/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar/calendar-custom-tab.js
+++ b/Jellyfin.Plugin.JellyfinEnhanced/js/arr/calendar/calendar-custom-tab.js
@@ -93,7 +93,12 @@
       //    content for tabs the user never opened.
       var tabContent = el.closest('.tabContent');
       if (tabContent) {
-        if (tabContent.classList.contains('is-active')) return el;
+        // `is-active` is the fast path but is not authoritative on its own:
+        // not every host sets it, and it can be applied without a DOM
+        // mutation for the observers below to notice. An actual visibility
+        // check backs it up so the panel still mounts if the class is absent
+        // or lands late — a missing mount is far worse than a late one.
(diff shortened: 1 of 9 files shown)


4eh5xitv6787h645ebv:
I have a feeling this leaves broken pages. I'll have to check in a few hours when I'm home.

4eh5xitv6787h645ebv:
Okay confirmed this would leave blank tabs. / non loading ones. Also reintroduces bugs where a different user can see another's on an account switch.

0-zach:
Hey, thanks for catching those.

The blank tab was me gating on `is-active`. The body observer only fires on added/removed nodes, so it can't see that class change at all. I added a visibility fallback plus a scoped observer on the panels' parent, so activation should always mount now.

For the user switch, the modules were relying on a re-render to refresh what's on screen and a tab that isn't open never gets a re-render. They should now clear the rendered DOM on the transition too, not just the in-memory state.

Unfortunately, I wasn't able to reproduce the user change bug on my setup, nothing survived sign-out. Instead, I tested this by firing `je:user-data-loaded` directly, and with the tab closed the content clears and stays cleared.

Thanks for taking a look, and please let me know if there's anything else.

4eh5xitv6787h645ebv:
tabs can remain blank after navigation. hmm maybe wait for an update to the custom tabs plugin first. 

also cant replicate it slowing the home page down anyways. As it loads aysnchronously. So home will not be slowed down by the calendar waiting on arrs to finish. Unless you can provide something that says otherwise?

0-zach:
Hey, you're completely right about the fetch being async. The fetch doesn't block the home render, so "slowing the home page down" was overstated. What I can say is that it triggers an uncached call out to Sonarr/Radarr when the home page mounts it, for a tab that was never opened, which is wasteful but not what I opened with.

On the blank tabs, I think I found the reason. Jellyfin's view cache can keep two `#indexPage` instances alive at once, and both calendar panels report `is-active: true`:

```
#0  kids=1  act=true  hidden=false  vis=true    <- live
#1  kids=1  act=true  hidden=true   vis=false   <- stale
```

`findActiveContainer()` walks the list backwards, so it can return the stale container inside the hidden page and the visible tab gets nothing. Document order isn't stable either: the live page was #1 in one of my tests and #0 in another. For what it's worth the existing code has the same hole, it's just masked by re-mounting on every mutation.

I completely agree this is better solved in the custom tabs plugin than worked around here, so I'll close this out. Thanks for the reviews.

4eh5xitv6787h645ebv:
Oddly enough I only just randomly ran into this issue. Only happens to me with Firefox for some reason. Not chromium.. I see there's a fix already inupstream for it.

PR #812, opened 2026-09-08
Revert #798: restore protected request avatar proxying

4eh5xitv6787h645ebv:
Reverts #798 (merge commit db78bee9fd126e15ff8e647b1e9f930289c07e0e), restoring the authenticated avatar proxy for every non-empty requester avatar returned by `GetRequests()`.

#798 returns any avatar beginning with `https://` directly to the browser. When that requester appears on a request card, the browser contacts the supplied host and bypasses the proxy's content-type checks and caching. This revert restores the previous behavior: relative Seerr avatars load through the proxy, while absolute URLs are rejected before an upstream avatar fetch.

**Known tradeoff:** this also restores the original local-user Gravatar limitation. Absolute Gravatar URLs receive HTTP 400 from the proxy and their avatars remain hidden. This PR is an exact, one-file revert; it does not add Gravatar proxy support or change avatar handling on other surfaces.

Verification completed with a controlled Seerr API fixture and isolated Docker containers:

- Release builds for `JellyfinTarget=jf10` and `jf12`: zero warnings and errors.
- Live reverted plugin on Jellyfin **10.11.11** and **12.0-rc6**, compared with current `main` on 12.0-rc6.
- Chromium using the actual request-card renderer and avatar helpers: the baseline attempted three external avatar requests on initial render; the revert attempted **zero**, including after rehydration. External requests in the baseline test were intercepted locally.
- `/avatarproxy/`, `/avatar/`, and `/api/v1/avatar/` images load; client cache reuse, server cache reuse, and ETag/304 responses pass.
- Arbitrary HTTPS, mixed-case HTTPS, Gravatar, traversal, and protocol-relative inputs are blocked; SVG/HTML responses are rejected; empty/missing avatars remain absent. Both endpoints still require authentication.
- No JavaScript exceptions in the browser harness; `git diff --check` passes.

Developed and verified with Codex assistance. Browser verification used real plugin modules and live plugin endpoints with fixture data, rather than a full Seerr installation.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Controllers/JellyfinEnhancedController.cs
@@ -9244,22 +9244,11 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Controllers
                         string? username = requestedBy?["username"]?.Value<string>();
                         string? avatar = requestedBy?["avatar"]?.Value<string>();
 
-                        // Proxy avatar through our backend to avoid CORS/mixed content issues.
-                        // Jellyseerr returns two different shapes in `user.avatar`:
-                        //   - a relative path for media-server users, e.g. "/avatarproxy/<jellyfinUserId>"
-                        //   - an absolute Gravatar URL for locally-created Jellyseerr users,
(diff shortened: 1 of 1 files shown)


n00bcodr:
FYI @Kisnov

Kisnov:
sorry for the trouble!

PR #824, opened 2026-09-12
fix(analytics): refresh reports after build or host changes

4eh5xitv6787h645ebv:
Correcting a Jellyfin Enhanced installation from jf10 to jf12 at the same plugin version currently leaves analytics unchanged until the normal reporting interval elapses. Upgrading Jellyfin alone has the same problem because the due-check only compares the first three components of the plugin version.

Compare the full plugin assembly version, compiled build target and running Jellyfin version against the last successful report. Any change triggers a report at the next startup/scheduled check. Older configurations refresh once to populate the missing fields. Failed sends leave the previous combination intact so the next check retries; analytics opt-out still applies.

Store the full version locally while preserving the backend's existing three-component payload format. Include the new fields in configuration-state preservation and regenerate the documentation defaults. This improves reporting freshness; it does not replace an incorrectly installed plugin DLL.

Validation:
- Both jf10 and jf12 plugin builds pass with zero warnings or errors.
- 15 local offline regression cases pass, compiling the complete production reporting service against dependency stubs and an intercepted HTTP handler. Covers build/host/revision changes, legacy configurations, cadence, successful persistence, failed-send retries, opt-out and configuration replacement. The harness is outside this PR; no live Jellyfin/backend integration test was performed.
- `git diff --check` passes.

Follow-up to #799. AI-assisted implementation and validation.

### DIFF
--- a/Jellyfin.Plugin.JellyfinEnhanced/Configuration/PluginConfiguration.cs
+++ b/Jellyfin.Plugin.JellyfinEnhanced/Configuration/PluginConfiguration.cs
@@ -395,6 +395,8 @@ namespace Jellyfin.Plugin.JellyfinEnhanced.Configuration
             AnalyticsLastReportedAt = 0;
             AnalyticsLastPayloadJson = string.Empty;
             AnalyticsLastReportedPluginVersion = string.Empty;
+            AnalyticsLastReportedJellyfinTarget = string.Empty;
+            AnalyticsLastReportedJellyfinVersion = string.Empty;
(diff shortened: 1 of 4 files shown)


4eh5xitv6787h645ebv:
Would be the only reason some installs still show wrong in the community anayltics.

n00bcodr:
> Would be the only reason some installs still show wrong in the community anayltics.

Hopefully!

Thanks as always!

Issue #827, opened 2026-09-12
🐛[BUG] Remove from Continue Watching succeeds but leaves the action sheet open


quantumbagel:
Note: this is my first attempt at contributing to this project so let me know if there's anything I can do. I will also open a PR for this issue as I do have a working fix.

It is somewhat confusing to me that this is such a reproducible issue, but here we go :)

**Describe the bug**
Clicking *Remove from Continue Watching* (or *Remove from Next Up*, which go to the same code) in an item's `⋯` action sheet successfully hides the item, but the action sheet stays open. Other native items in the same menu close it with the normal scaledown animation.

The Remove handler calls `preventDefault()` / `stopPropagation()`, so jellyfin-web never runs its normal close path. After the hide POST succeeds, `closeOpenActionSheet()` tries to dismiss the sheet itself with `HTMLDialogElement.close()` and a synthetic Escape `keydown`. This doesn't seem to work and leaves the menu on screen.

**Expected behavior**
After clicking Remove, the action sheet should close the same way every other item in that menu does. The card should disappear from Continue Watching / Next Up.

**Screenshots**
N/A

**Jellyfin Environment (please complete the following information):**
 - Server Version: 12.0 (Although note that I've been having this issue since I installed jellyfin a few months ago w/ version 10.11.11)
 - Plugin Version: 12.7.0.0
 - Jellyfin Installation Method: Docker
 - Device: Jellyfin Web
 - External Access Method: Caddy
 - Theme being used: ElegantFin (The bug is reproducible on my end even without a theme enabled)
 - [x] [FileTransformation](https://github.com/IAmParadox27/jellyfin-plugin-file-transformation) Installed

**Logs**
There is no server error I can see pretty much anywhere. The hide request succeeds and does actually remove the media. It is purely a visual thing.

**Additional Context**

The way I fixed it locally:
* Let the click work naturally instead of blocking it.
* Remove `preventDefault()`, `stopPropagation()`, and `closeOpenActionSheet()` from the Remove button handler.
* Let the click bubble up so Jellyfin-web can close the menu automatically, just like it does for Play or Mark Played.
* Keep all of the existing code that actually implements the logic. It can finish running in the background while the menu closes. I can change the behavior to wait if that is wanted but I don't particularly think it matters too much?

let me know if there's any other context/files that are needed :)

PR #828, opened 2026-09-12
fix: continue watching/next up removal not closing

quantumbagel:
See #827


4eh5xitv6787h645ebv:
Does this break the tv mode(jellyfin webui tv mode) of remove cw?

quantumbagel:
Nope! I hadn't tested this, but glad it seems to be working :)

Video of the button working in TV mode (disabled theme as well for the recording):

https://github.com/user-attachments/assets/bf8d425e-1625-4003-aba0-41b5def503be

PR #845, opened 2026-09-21
fix: defer inactive Home tabs and stop stale Activity polling

4eh5xitv6787h645ebv:
When Home is visible, the tab mount checks accept its page before checking whether the enclosing tab is selected. This renders inactive Calendar, Downloads, Hidden Content, Recommendations, Activity and Bookmarks content, and can leave Downloads polling in the background. Random also waits for a later DOM mutation before its initial insertion.

Require every enclosing page/tab to be active and observe their class/hidden changes through a shared, disposable watcher. It uses the existing after-paint scheduler, including its background-tab fallback. Bookmarks reconciles replaced or cleared containers after an in-flight render and invalidates cached content when bookmarks change, and Activity invalidates pending requests when polling stops so stale completions cannot restart or cancel a newer poller. Random is added during initial UI setup. Native tab panels follow rebuilt Home pages, including pages inserted hidden and later revealed; an applied deep link no longer overrides subsequent user tab selections.

Based on current upstream `main` (`0ae2e2a0`), preserving its body-observer deferral and batched visibility reads. This PR contains startup/tab mounting only; the single-row header change is covered separately in [#844](https://github.com/n00bcodr/Jellyfin-Enhanced/pull/844). Both changes were also validated together on the installed Jellyfin 12 test server.

## Relationship to the existing performance fix

This builds on [ee9b0b0c](https://github.com/n00bcodr/Jellyfin-Enhanced/commit/ee9b0b0c198c71e5a901229f92c69c13b6cb6666), already included in the base. That commit batches shared observer subscribers after paint and reduces forced layouts. This PR preserves that scheduler and visibility cache; it prevents inactive tab content from mounting in the first place, handles activation/remounts, and stops stale polling. It does not claim the upstream commit's reported freeze reduction or a separately measured wall-clock startup speedup.

The current-upstream browser fixture still starts all five tested inactive tab renderers; the same startup assertion passes with this patch (zero inactive renders). Live tests also show the five configured panels remain unrendered until selected.

## Validation

- Local browser regression checks pass in Chromium and Firefox (test scripts are excluded from this PR). The startup regression fails against unchanged upstream. Cases cover inactive/nested/hidden tabs, class-only activation, suspended animation frames, disposal, the real Bookmarks renderer with pending requests, Activity polling races, native remounts and deep links. Historical blank-tab coverage includes replacement of `.mainAnimatedPages` after dashboard navigation (issue 536), layouts without a `.page` wrapper (issue 570), hidden placeholders, pending bookmark requests, user-data changes and class-only page reveals. The two new blank-tab tests fail on the previous PR revision and pass with these corrections.
- Live Jellyfin 10.11 and 12 checks in both browsers: five configured inactive panels remain unrendered on Home, Random is mounted, every tab loads when selected, Downloads stops polling on Home, and direct tab URLs/remounts work. Two dashboard/Home round trips per browser/version verify all five configured tabs still render.
- Release builds for `jf10` and `jf12` pass with zero warnings/errors. Changed JavaScript syntax checks and `git diff --check` pass.

Existing native-controller console errors on Jellyfin 10 tab selection and cold direct-tab URLs also reproduce on unchanged upstream; this PR does not claim to resolve those loader errors. Ordinary Home startup produced no page exceptions in the live checks.

Developed, reviewed and tested with AI assistance (OpenAI Codex).


4eh5xitv6787h645ebv:
Didn't realize you already had a fix for it 🤣

This was the way iwas going about it.

Also I could only get this to happen on Firefox. Not Chromium. Did you have the same thing happen or nah?

n00bcodr:
I only use Chrome, never did see a freeze, although with [ee9b0b0c](https://github.com/n00bcodr/Jellyfin-Enhanced/commit/ee9b0b0c198c71e5a901229f92c69c13b6cb6666) I have seen a huge improvement!

